Issue #1, opened 2016-12-13
Automate Test Creation


lithp:
- [x] Create cloudformation template which only spins up a cluster
- [x] Decide whether to use bash scripts or puppet / fabric
- [x] Create happy-path script which only sets up PG and Citus to enable documentation testing (it should accept which version of PG and which branch/commit of Citus to use)
- [x] Add the ability to specify additional files to include and commands to run on the citus cluster (like creating and staging tables). Write a target for TPCH testing.
- [x] Add the ability to build pg/citus with custom configure flags (--enable-debug) and write a target for the Valgrind testing
- [x] add the ability to specify which additional extensions are installed and create a target for integration testing (which wants session analytics and shard rebalancer)
- [x] add the ability to specify which additional programs are installed and create a target for feature testing (which wants JDBC)
- [ ] Write the sample queries DML testing should use and a target for it.
- [x] add the ability to have multiple simultaneous versions of pg and citus installed in the target, then create an upgrade testing target
- [x] Write a target for citus enterprise testing
- [x] Between versions of Citus the steps to setup a cluster change, how do you accommodate that? Likely by adding a setup file to the Citus repo?
- [ ] Automate Integration Testing: Create a script which runs the citus, session analytics and shard rebalancer regression tests, and an additional regression test to test behavior, then add that script to Travis (which first sets up the integration testing target).

Onder and the rest of the team had some feedback on the order these should be done in, which I haven't yet incorporated into the above list.

(Solves https://github.com/citusdata/citus/issues/989)

lithp:
The first phase has been completed and merged

PR #10, opened 2016-12-16
The initial PR

lithp:
Will eventually contain the results of checking all these boxes: https://github.com/citusdata/test-automation/issues/1


README.md
```diff
@@ -1,2 +1,83 @@
...
+# Example fab commands
+
+Once you have a cluster you can use many different variations of the "fab" command to
+install Citus. Here are some examples:
+
+- `fab --list` will return a list of the tasks you can run.
+- `fab setup.basic_testing`, will create a vanilla cluster with postgres and citus. Once this has run you can simply run `psql` to connect to it.
+- `fab use.citus:v6.0.1 setup.basic_testing` will do the same, but use the tag `v6.0.1` when installing Citus. You can give it any git ref, it defaults to `master`.
+- `fab use.postgres:9.6.1 setup.basic_testing` lets you choose your postgres version.
+- `fab use.citus:v6.0.1 setup.enterprise` will install postgres and the `v6.0.1` tag of the enterprise repo.
```
onderkalaci: I think there is a lot more to explain in the `README` which could dramatically decrease people asking questions to you over and over :) 

* We should briefly mention what do  `setup`, `use` and `pg` mean for us. In that way, users could comprehend the framework more easily.
* We should be mentioning `valgrind` setup. There is no single word about it.
* We should be mentioning `shard_rebalancer` once we implement it.
* Should show an example or a more detailed explanation of using a closed source project. I wasn't able to run them with following the README. For instance, `citus-enterprise` and `rebalancer` installed on a single cluster.
* `psql postgres` connects to another database. So, should emphasize to simply use `psql` (because, I did that couple of times and got confused)

fabfile/add.py
```diff
@@ -0,0 +1,92 @@
...
+        run('git checkout {}'.format(git_ref))
+        run('make install')
+
+    # TODO: What if the server isn't running?
+    with cd(prefix):
+        run('bin/psql -c "CREATE EXTENSION session_analytics;"')
+
+@task
+def hll():
+    'Adds the hll extension to the instance in pg-latest'
```
onderkalaci: Minor: What about moving function comments to just before the function definition as we do on Citus repo?

cloudformation/find-workers.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh -e
+
+if [ -z "$AWS_DEFAULT_REGION" ]; then
+    export AWS_DEFAULT_REGION=$(curl -s http://169.254.169.254/latest/dynamic/instance-identity/document | grep region | awk -F\" '{print $4}')
```
lithp: @marcocitus would know best but I think the answer is no. The documentation suggests using the ip address http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html

With the level of control AWS has over the network, this ip is basically a dns name ;)

README.md
```diff
@@ -1,2 +1,83 @@
 # test-automation
-Tools for making our tests easier to run
+
+Tools for making our tests easier to run. Automates setting up a cluster with
+CloudFormation and installs a script which automates setting up citus and everything
+required for testing citus.
+
+# Starting a cluster
+
+You'll need to have installed the (https://aws.amazon.com/cli/)[AWS CLI]. Once that's
```
lithp: I think everyone has `ssh-agent` installed. And if we get Windows developers in the future `ssh-agent` will be the least of their problems :)

I'll look at the broken link and potentially inlining some instructions though.

README.md
```diff
@@ -1,2 +1,83 @@
...
+
+- `fab use.citus:v6.0.1 setup.enterprise add.shard_rebalancer` does what you'd expect.
+
+# Advanced fab usage
+
+By default your fab commands configure the entire cluster, however they can be targeted
+at roles or individual machines.
+
+- `fab -R master pg.restart` will restart postgres on the master node
+- `fab -R workers pg.stop` will shutdown pg on all the workers.
```
lithp: Yeah, that's a great idea, we can easily ask cloudformation to add these to `/etc/hosts`.

cloudformation/upload-template.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
lithp: This was a helper script for my own development. It's probably worth deleting.

fabfile/add.py
lithp: These comments (docstrings) are automatically grabbed by `fab --list` and are the way documentation works in Python. Moving them to before the function definition would make `fab --list` far less useful and isn't very pythonic at all.

onderkalaci: Wow, that's nice :)

cloudformation/find-workers.sh
marcocitus: Yep, this is a link-local IP address that the EC2 host machines use for making metadata available to EC2 instances. It will never change.

README.md
```diff
@@ -1,2 +1,83 @@
...
+Finally, there are `add` tasks, such as `add.tpch`. These assume that a cluster is already
+installe and running (and therefore must come after the `setup` tasks, and install
+additional components, such as the tpch data.
+
+# `add` tasks
+
+It is possible to add extra extensions and features to a Citus cluster:
+
+- `fab add.tpch:scale_factor=1` will generate and stage tpch tables (the default scale factor is 10)
+- `fab add.session_analytics` will build and install the session_analytics package (it uses your ssh agent forwarding when connecting to github, so you must have the key you use for github added to your local agent)
```
marcocitus: You could add the public key of your EC2 key pair to https://github.com/settings/keys I think.

cloudformation/create-stack.sh
```diff
@@ -0,0 +1,175 @@
...
+        # Print error messages if any
+        echo >&2
+        aws cloudformation describe-stack-events \
+            --stack-name ${stack_name} \
+            --query "StackEvents[?ResourceStatus==\`CREATE_FAILED\`].ResourceStatusReason" \
+            --output text
+
+        echo >&2
+        echo Deleting stack ${stack_name} >&2
+        aws cloudformation delete-stack --stack-name "${stack_name}"
```
lithp: Here's the command: `aws cloudformation list-stacks --stack-status-filter CREATE_COMPLETE CREATE_IN_PROGRESS CREATE_FAILED`. I'll add it to the README.

lithp: Also, created [a ticket](https://github.com/citusdata/test-automation/issues/24) for the error you experienced.

fabfile/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+import config # settings, but no tasks
+import pg  # tasks which control postgres
+import add  # tasks which add things to existing clusters
+import setup  # tasks which create clusters with preset settings
+import use  # tasks which change some configuration future tasks read
+import run
+
+@task(default=True)
+@runs_once
+def main():
```
lithp: I did this because the `create-stack.sh` script runs `fab` automatically when it can and I didn't want to have to hardcode the name of the task to run, in case the appropriate task changes in the future.

I'm a little worried about maintaining all this documentation. If I follow all your comments I'll have to document three places:
- The docstrings of the tasks, to make sure `fab --list` works.
- The README, to list all of the tasks and what they do again.
- The default `fab` task.

Keeping all of these in sync will be a bit tedious.

README.md
```diff
@@ -1,2 +1,91 @@
 # test-automation
-Tools for making our tests easier to run
+
```
onderkalaci: I'm summarizing the items that we should be fixing before merging in this:

* [ ] Make sure that enterprise installation works as expected
    * [ ] The branch is manually coded
    * [ ] Output of the following command is confusing, might be a bug there since we're cloning from `citus` repo instaed of enterprise:
```
fab use.citus:v6.0.1 setup.enterprise
[localhost] Executing task 'use.citus'
[localhost] local: rm -rf /home/ec2-user/citus || true
[localhost] local: git clone -q https://github.com/citusdata/citus.git /home/ec2-user/citus
```

* [ ] Make sure that shard rebalancer can be installed. Currently, there is no tasks associated with it.

* [ ] Make sure to document @marcocitus 's comment about key-pair issue on README regarding closed source projects. That worked for me.

* [ ] Make sure that `cstore_fdw` installation updates `postgresql.conf`. It'd be awesome if you create a generic `update_postgres_config` task that we could use for other purposes as well. I had some suggestions on how to do that in the PR.

* [ ] Make sure that postgresql source is complied with `-j ` option. That could decrease the compilation time dramatically.

* [ ] Enable running regression tests for `citus` and `postgres` with a task. And, document them.

* [ ] Try to check in #23. That could save a lot of time for us.


After tackling the above items, could you ping me so that I can do a final testing?

Thanks!

lithp: I checked in most of these. Still need to update the README with better instructions but it might be ready for another look.

lithp: Updated the README with more thorough instructions for connecting to the master and messing with ssh-agent

README.md
```diff
@@ -1,2 +1,83 @@
...
+By default your fab commands configure the entire cluster, however they can be targeted
+at roles or individual machines.
+
+- `fab -R master pg.restart` will restart postgres on the master node
+- `fab -R workers pg.stop` will shutdown pg on all the workers.
+
+You can also ask to run arbitrary commands by adding them after `--`.
+
+- `fab -H 10.0.1.240 -- cat "max_prepared_transactions=0" >> pg-latest/data/postgresql.conf` will modify the postgresql.conf file on the specified worker.
```
lithp: Added `pg.read_config:key` and `pg.set_config:key,value`.

README.md
onderkalaci: Hey Brian,

I think this looks pretty good now. This is a conditional :shipit: , once you fix the below items, feel free to merge to the master:

- [ ] Make sure to squash into a single (or a few) commits and have a meaningful commit message.
- [ ] Citus installation that are pre 6.0 does not have `master_add_node()`. So, for instance `tpch` tests do not work with them. Could we add workers to `pg_worker_list.conf` for pre-6.0 citus installations? (Well, you integrated that change to Citus from MX, so I think you could remember how to do pre 6 installations😄   ) I think there is no harm of adding pg_worker_list.conf for citus >= 6 installations, so maybe add to all installations to make it simple?
- [ ] tpch scale factor does not work:
```
ec2-user=# \q
[ec2-user@ip-10-192-0-174 ~]$ fab add.tpch:scale_factor=1
[localhost] Executing task 'add.tpch'
[localhost] run: readlink /home/ec2-user/pg-latest
[localhost] out: /home/ec2-user/pg-950-citus-522
[localhost] out: 

[localhost] run: make
[localhost] out: chmod 755 update_release.sh
[localhost] out: ./update_release.sh 2 13 0
[localhost] out: 

[localhost] run: SCALE_FACTOR=10 CHUNKS="o 24 c 4 P 1 S 4 s 1" sh generate2.sh
```

- [ ] Well, multiple installation capability is really neat. But, using it is not very intuitive. Please add your thoughts to #30 on how could we make it more usable? If you don't have any, please feel free to skip this.

- [ ] I couldn't see regression tests for postgres. I'm OK with not implementing/fixing this with the condition that we open an issue for that 😄 

- [ ] I tried to set preload_shared_libraries config, which I failed `fab pg.set_config:shared_preload_libraries,'citus, cstore_fdw'`. Is there a way to run this? If so, could you document it? If not, could you open an issue for that?
- [ ] In the `Ordering of fab tasks`, could you please give an example such that we could build both asserts and debug mode enabled? I mean, it is great that we have the explanation of what to do, but, I'd also want to simply copy&paste a working command with 
- [ ] Running `fab add.cstore` does not add `cstore_fdw` to preload shared libraries until `fab pg.restart`. Could we make that step automatic? 
- [ ] Due to some mistake or intentional changes, I got stuck such that `fab pg.stop` or `fab pg.restart` was erroring out saying that are you sure postgres is running. I'd really want to see that when I say `fab pg.restart`, the instances are restarted irrespective of being stopped. Believe me, I hit this quite a few times on `multiple installation` setup. And, I'm pretty sure that people will hit this a lot.

sumedhpathak:
@onderkalaci @lithp What needs to be completed here before we can merge? Are the changes ready to be used for 6.1 testing?

cloudformation/create-stack.sh
```diff
@@ -0,0 +1,175 @@
+#!/bin/bash -e
```
lithp: Marco fixed this, it deletes the cluster on exit now

lithp:
I ticked off all the boxes and created issues for any feedback which wasn't marked as required.

PR #13, opened 2016-12-16
Create symbolic link to fabfile.py instead of copying as root

marcocitus:
By copying the new file is owned by root, might be nicer to just link it.


lithp:
Nice, thanks!

PR #14, opened 2016-12-16
Make a pg-latest symbolic link and add pg-latest/bin/ to PATH

marcocitus:
Just so we can type `psql` right after running `fab basic_testing`.


lithp:
Sounds good to me

PR #16, opened 2016-12-21
Improve error handling and automate basic_testing set-up

### LINKED ISSUE #15: Worker node count flag didn't work
I tried to create 3 workers, but got 2.

marcocitus:
Fixes a few issues with the `create-stack` script, in particular:

- Fixes #15 (make `--num-workers`/`-n` work) 
- If you specify `--private-key-file` (`-p`), the script will automatically run fab basic_testing (currently fails with "Fatal error: Couldn't find any fabfiles!" because of dead symbolic link)
- If the CloudFormation stack fails to create, it shows the CREATE_FAILED error message
- Fail early if specified files don't exist
- Automatically find a citus.json file in the directory


lithp:
Sorry about the dead link, that should be fixed now. I'll test this out soon, all of these changes make sense.

PR #50, opened 2017-01-31
Add tpch hash suite

metdos:
Example usage:

```
fab add.tpch:scale_factor=1,partition_type='hash'
```

### DIFF
--- a/tpch_2_13_0/tpch_create_hash_partitioned_tables.ddl
+++ b/tpch_2_13_0/tpch_create_hash_partitioned_tables.ddl
@@ -1,0 +1,116 @@
+-- Sccsid:     @(#)dss.ddl	2.1.8.1
+
+CREATE TABLE nation
+(
+    n_nationkey  INTEGER not null,
+    n_name       CHAR(25) not null,
+    n_regionkey  INTEGER not null,
+    n_comment    VARCHAR(152)
+);
+
+select create_distributed_table('nation', 'n_nationkey');
+
+
+CREATE TABLE region
+(
+    r_regionkey  INTEGER not null,
+    r_name       CHAR(25) not null,
+    r_comment    VARCHAR(152)
+);
+
+select create_distributed_table('region', 'r_regionkey');
+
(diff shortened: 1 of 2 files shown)


lithp:
Nice, looks good to me!

onderkalaci:
Needs documentation 😄

tpch_2_13_0/tpch_create_hash_partitioned_tables.ddl
```diff
@@ -0,0 +1,116 @@
...
+
+CREATE TABLE nation
+(
+    n_nationkey  INTEGER not null,
+    n_name       CHAR(25) not null,
+    n_regionkey  INTEGER not null,
+    n_comment    VARCHAR(152)
+);
+
+select create_distributed_table('nation', 'n_nationkey');
```
onderkalaci: I think this should be `create_reference_table()` Same for `supplier `. @metdos Does that work?

metdos:
> I think this should be create_reference_table() Same for supplier. @metdos Does that work?

Yeap that makes sense. Just, it would be conflicting with the settings, because we say 'hash' but also create reference tables.

We can provide 3 options;
1. Default, hash + reference tables.  
2. 'hash' -> all tables are hash partitioned
3. 'append' -> all tables are append partitioned

lithp:
What does having multiple options save us? Why would we want only hash tables? And why do we want a mix of hash+references tables?

metdos:
> What does having multiple options save us?

When you need them, it saves us from creating them manually. Same as why we need test automation.

> Why would we want only hash tables?

I would want to run some tests with only hash tables. You could not replicate this problem otherwise:
https://github.com/citusdata/citus/issues/542

> And why do we want a mix of hash+references tables?

Because some of them need to be reference tables in real life scenarios so it is the default. Other than that it is a good opportunity test them together.

lithp:
Okay. It feels a little weird to provide all three options but if all all of them are independently useful I suppose it's not a lot of extra cruft.

PR #52, opened 2017-02-01
Change some tpch tables to reference table

onderkalaci:
`nation` and `supplier` tables are reference tables in the regression tests.


metdos:
Closing regarding #53

PR #63, opened 2017-09-07
task executor changes and hardcoded variable removals

furkansahin:


jdbc/JDBCReleaseTest.java
```diff
@@ -7,10 +7,11 @@
 import java.util.Random;
 
 public class JDBCReleaseTest {
-	
-	static String url = "jdbc:postgresql://localhost:5432/ec2-user";
-	
-	
+
+	private static final String user_name = "USER_NAME";
+	static String url = "jdbc:postgresql://localhost:9700/postgres";
```
mtuncer: This does not look right.

It requires this file to be edited when running on automation to fix the url and user name.

I would change url to ```"jdbc:postgresql://localhost:5432/postgres"``` and fetch user name from the current user logged on the system.

```java
System.getProperty("user.name");
```

should return currently logged in user name.

SaitTalhaNisanci:
Should we close this PR?

PR #69, opened 2017-10-31
Add usability fixes

velioglu:
Fix some usability issues and let PG 10 can be used


metdos:
@onderkalaci @velioglu merging this to move fast. In the upcoming days, I will look to generalize DML and tpch tests, so we can use them for a greater set of tests.

PR #70, opened 2017-11-01
Fix usability errors and add PG 10 support

velioglu:
Fixes error against PG 10 and improve usability of fab


onderkalaci:
What happens if I write `onderTest` to Postgres version now?

velioglu:
Installation process will be error out due to 404 error.

furkansahin:
PR looks good to me

PR #82, opened 2017-12-29
Fix set_config

metdos:


metdos:
Thanks to @onderkalaci for reporting...

PR #87, opened 2018-01-25
Add support for PostgreSQL in the main TPC-H files.

DimCitus:
This now has quite some amount of work with no hope to easily sync with TPC-H ever, as this PR contains the full support for the _direct loading_ of the data to PostgreSQL, using the COPY protocol without intermediary files. It might be too invasive for this repository, and we could maybe have that _vendor fork_ of the TPC-H files live elsewhere.


metdos:
You don't need to use citus_indent here. Also, FYI, we don't use citus_indent on PostgreSQL files like rule_utils which makes integration to newer versions harder.

tpch_2_13_0/qgen.c
```diff
@@ -205,18 +210,23 @@ char *cptr,
                     break;
                 case 'n':
                 case 'N':
+				{
                     if (!(flags & DFLT_NUM))
-                        {
+					{
```
metdos: Such as this change makes it a bit harder to read.

metdos:
I liked the changed output, it would be better if you revert back indentation changes to see modifications better.

DimCitus:
I made more changes and I don't think it's a good idea to publish them here, I'm going to handle that in my own tpch repository with more scripting around it. We can always change our mind later of course.

metdos:
Sounds good!

PR #90, opened 2018-05-24
fabric update fix

onderkalaci:
We've hit some version/API incompatibilities since the default version of fabric has changed. We propose to fix as in this PR by explicitly setting the version.

@lithp Do you think is this OK?


lithp:
Yeah, sounds good to me!

lithp (approved):

lithp:
Probably a better solution going forward is to use the newer version of Fabric and change our framework to work well with it.

PR #94, opened 2019-09-02
Refactor home directory

SaitTalhaNisanci:
This PR:

- Uses home directory from config instead of hard coded `/home/ec2-user`
- Adds `pyc` and `vscode` to `.gitignore` 

All of the paths in https://github.com/citusdata/test-automation/blob/master/cloudformation/citus.json are also hard coded, this should also be refactored. If we move to azure cloudformation will not be used so we can do this refactor later.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+*.pyc
+.vscode/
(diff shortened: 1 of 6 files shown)


metdos:
If you tested this branch, you can merge it.

PR #97, opened 2019-09-09
Add basic citus upgrade tests

SaitTalhaNisanci:
This PR adds simple upgrade tests, it checks:
- Upgrading citus version is successful with server restart
- Upgrading citus version without restarting server forces you to restart server

To run the test run :
```
fab run.update
```


fabfile/constants.py
```diff
@@ -0,0 +1,3 @@
+DROP_CITUS = 'DROP EXTENSION IF EXISTS citus;'
+CREATE_CITUS = 'CREATE EXTENSION citus;'
+ALTER_CITUS = 'ALTER EXTENSION citus UPDATE;'
```
hanefi: If we do not provide `TO version` at the end, this will update to the latest version available. Is this really what we want?

hanefi: and maybe have aline ending at the end as well

fabfile/run.py
```diff
@@ -198,3 +199,42 @@ def tpch_queries(query_info, connectionURI, pg_version, citus_version):
         out_val = run(run_string)
         results_file.write(out_val)
         results_file.write('\n')
+
+@task
+@runs_once
+@roles('master')
+def update():
+     citus_versions = ['release-8.0', 'release-8.1']
+     pg_version = '11.5'
```
hanefi: similarly, instead of a hardcoded value, can this be a parameter as well?

hanefi: An alternative solution is to use `config.settings['pg-version']`
This way we can override this by adding some initial use commands. For example

	fab use.postgres:12.0 run.update:8.1:8.2

hanefi (changes requested):
I did not test the changes here myself yet. Left some initial reviews. Looks good overall

SaitTalhaNisanci:
@hanefi we know have citus upgrade tests in our CI, I think we can close this, I dont see any benefit of this PR now, what do you think?

SaitTalhaNisanci:
Closing this since we have these tests as part of our CI jobs now.

PR #99, opened 2019-10-01
Update test configs with recent PG and Citus versions and use enterprise

SaitTalhaNisanci:
This PR updates all test configs to use pg11.5 and citus release-9.0.
It also changes all tests to use citus enterprise repository.

We can possibly change the `postgres-citus-versions` from a single variable instead of having them in each file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,8 +85,9 @@ ssh -A ec2-user@ec2-35-153-66-69.compute-1.amazonaws.com
 ```
 
 On the coordinator node:
+
 ```bash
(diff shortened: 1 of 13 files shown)


SaitTalhaNisanci:
When we are doing these tests for release-x, we do not have release-x branch in citus community repo. So if we change configs which has `use enterprise: off`, they will fail. Is that right?

metdos:
> We still seem to have some old configuration files. i guess we can update all of them

These are examples, and we don't run them in every release, so I'm ok to keep them as it is.

SaitTalhaNisanci:
Well I have updated all of them now

PR #103, opened 2019-10-10
Add github to known hosts in use enterprise

SaitTalhaNisanci:
We first `git clone` to see if the given repo exists, however for cloning enterprise, it prompts if we want to trust `git`. We should add github to known hosts there to bypass that.


hanefi (approved):
:shipit:

PR #111, opened 2019-10-14
put config name to results file name and use separate results dir

### LINKED ISSUE #106: More descriptive result file names
When we run a test, we should put the config file name to that test's result file name, so that it is easier to understand what config was used for getting that result.

### LINKED ISSUE #107: Separate results folder
We should put our results to a separate folder so that it is easier to read/copy them.

SaitTalhaNisanci:
Fixes #107.
Fixes #106.

### DIFF
--- a/fabfile/run.py
+++ b/fabfile/run.py
@@ -42,8 +42,11 @@ def pgbench_tests(config_file='pgbench_default.ini', connectionURI=''):
     config_folder_path = os.path.join(config.HOME_DIR, "test-automation/fabfile/pgbench_confs/")
     config_parser.read(config_folder_path + config_file)
 
+
     current_time_mark = time.strftime('%Y-%m-%d-%H-%M')
-    results_file = open(os.path.join(config.HOME_DIRECTORY, 'pgbench_results_{}.csv'.format(current_time_mark)), 'w')
+    utils.mkdir_if_not_exists(config.paths[config.RESULTS_DIRECTORY])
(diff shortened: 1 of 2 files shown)


fabfile/run.py
```diff
@@ -180,18 +183,19 @@ def tpch_automate(config_file='tpch_default.ini', connectionURI=''):
...
 @task
 @roles('master')
-def tpch_queries(query_info, connectionURI, pg_version, citus_version):
-    results_file = open(
-        os.path.join(config.HOME_DIRECTORY,'tpch_benchmark_results_PG-{}_Citus-{}.txt'.format(pg_version, citus_version)),
-        'a')
-    psql = '{}/bin/psql'.format(config.PG_LATEST)
-    tpch_path = '{}/tpch_2_13_0/distributed_queries/'.format(config.TESTS_REPO)
+def tpch_queries(query_info, connectionURI, pg_version, citus_version, config_file):
+    utils.mkdir_if_not_exists(config.paths[config.RESULTS_DIRECTORY])
```
JelteF: let's use https://docs.python.org/3.6/library/os.html#os.makedirs with `exists_ok` instead of writing the function ourselves (unless this a python2 project)

JelteF (changes requested):

fabfile/run.py
SaitTalhaNisanci: this is a python2 project for now.

PR #117, opened 2019-10-16
Add azure support

### LINKED ISSUE #115: Hard coded user name in scale test inputs
In scale test inputs, we have `ec2-user` as a hardcoded string.

### LINKED ISSUE #116: Add Azure support
Currently we are using cloudformation to create a cluster. We should support creating a cluster on Azure. It would be good to still keep the cloudformation in case we want to test there.

SaitTalhaNisanci:
Azure has ARM templates that can be used to deploy servers with ease. There are two main files for ARM templates, `azuredeploy.json` and `azuredeploy.parameters.json`. `azuredeploy.json` has the main template and `azuredeploy.parameters.json` contains the parameters that are used in the main template. For example if you want to change the number of instances, you would do that in the parameters. You shouldnt change anything in the template file for configuration.

  The main template has 4 main parts:
  
* Parameters
* Variables
* Resources
* Outputs

Parameters can be configured from the parameters file. Variables are constants. Resources have all of the resource definitions such as VMs, network security groups. Outputs can be useful for displaying a connection string.

When creating resources, we can specify the order so that if a resource depends on some other resource, it wont be created before the dependant is created. We can also specify how many instances of a resource to create with a `copy` command.

The first virtual machine with index 0 is treated as a coordinator. When all the virtual machines are ready, a custom script is installed to do initialization in vms. The initailization script is retrieved from the github with a url.

The initialization script also finds the private ip addresses of workers and puts them to the coordinator. The way this is done is with a storage account resource. This storage account resource is created within the template itself and all the vms upload their private ip addresses to the storage. After all are uploaded the coordinator downloads all the private ip addresses from the storage account and puts it to `worker-instances` file, which is then used when creating a citus cluster.

We have a special security group which blocks ssh traffic. The rule's priority is 103 and 100, 101, 102 are also taken by this security group. The workaround to connect to the coordinator is to remove the rule 103, and connect right after it. The rule comes back in 2-3 mins, so you should be fast here. There is a script called `delete-security-rule.sh`, which deletes that rule for you.

You can use `connect.sh` which will delete the security rule and connect to the coordinator for you.

Before starting the process you should set the environment variable `RESOURCE_GROUP_NAME`, which is used in all scripts.

```bash
export RESOURCE_GROUP_NAME=talhatest_resource_group
```

if you want to configure the region, you can also set that:

```bash
export AZURE_REGION=eastus2
```

You should use a single session because the exported variable is only available in the current session and its children sessions. You should start `ssh-agent` and add your key with `ssh-add`.

By default, your public key from `~/.ssh/id_rsa.pub` will be used. This public key will be put to the virtual machines so that you can ssh to them.

To simplify this process, there is a script called `create-cluster.sh`, which:

* creates a resource group from the environment variable `RESOURCE_GROUP_NAME`, and `AZURE_REGION`.
* creates a cluster with the `azuredeploy.json` template in the resource group
* prints the connection string to ssh

then you should run:

```bash
./connect.sh
```

or

```bash
./delete-security-rule.sh
ssh -A pguser@<public ip of coordinator>
```

After you are done with testing, you can delete the resource group with:

```bash
./delete-resource-group.sh
```

Currently the default time for tests 300 seconds, and as we have many tests it might take a while to run all the tests. So when testing a change, it is better to change the test times to something short such as 5 seconds. The time can be changed with the -T parameter:

```bash
pgbench_command: pgbench -c 32 -j 16 -T 300 -P 10 -r

->

pgbench_command: pgbench -c 32 -j 16 -T 5 -P 10 -r

```

If you want to add different vm sizes, you should change the allowed values for `coordinatorVMSize` and `workerVMSize` in `azuredeploy.json`.

Fixes #116.
Fixes #115.


onderkalaci:
For the readers, I suggested to keep ec2 code around for some time optionally, in case we want to run some benchmarks on ec2

README.md
```diff
@@ -26,7 +38,213 @@ required for testing citus.
...
+
+```bash
+
+# start ssh agent
+eval `ssh-agent -s`
+
+# Add your Github ssh key for enterprise (private) repo
+ssh-add
+
+# Put your public ip to sshPublicKey in azuredeploy.parameters.json. You can see your public key with:
```
SaitTalhaNisanci: This should be public key, not ip. This is a typo

azure/create-cluster.sh
```diff
@@ -0,0 +1,17 @@
...
+rg=${RESOURCE_GROUP_NAME}
+
+az group create -l eastus -n ${rg}
+
+az group deployment create -g ${rg} --template-file azuredeploy.json --parameters azuredeploy.parameters.json
+
+connection_string=$(az group deployment show -g ${rg} -n azuredeploy --query properties.outputs.ssh.value)
+
+echo "run './delete_security_rule.sh' to temporarily disable security rule, and connect with:"
+echo ${connection_string}
```
SaitTalhaNisanci: We should remove the quotes here. Currently it prints:

```bash
"ssh -A pguser@..."
```

we should have:

```bash
ssh -A pguser@...
```

azure/create-cluster.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+# fail if trying to reference a variable that is not set.
+set -u
+# exit immediately if a command fails
+set -e
+
+rg=${RESOURCE_GROUP_NAME}
+
+az group create -l eastus -n ${rg}
```
marcocitus: Can you change this to eastus2 or make it configurable?

azure/delete_resource_group.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
```
marcocitus: The create script uses dashes in the name and the delete script uses underscores. Should maybe standardize on dashes.

README.md
```diff
@@ -26,7 +38,213 @@ required for testing citus.
...
+ssh -A pguser@<public ip of coordinator>
+```
+
+After you are done with testing, you can delete the resource group with:
+
+```bash
+./delete_resource_group.sh
+```
+
+Currently each pgbench test is run for 300 seconds, and as we have many pgbench tests it might take a while to run all the tests. So when testing a change, it is better to change the tests times to something short such as 5 seconds. The time can be changed with the -T parameter:
```
marcocitus: This seems like it does not belong in the Azure section.

README.md
```diff
@@ -26,7 +38,213 @@ required for testing citus.
...
+```
+
+## <a name="azure-tpch-cloud"></a> Running TPC-H Tests Against Citus Cloud
+
+On the coordinator node:
+
+```bash
+# Provide your tpch config file or go with the default file
+# Don't forget to escape `=` at the end of your connection string
+fab run.tpch_automate:tpch_q1.ini,connectionURI='postgres://citus:dwVg70yBfkZ6hO1WXFyq1Q@c.fhhwxh5watzbizj3folblgbnpbu.db.citusdata.com:5432/citus?sslmode\=require'
```
marcocitus: Should this be updated for Hyperscale?

README.md
```diff
@@ -26,7 +38,213 @@ required for testing citus.
...
+./delete_resource_group.sh
+```
+
+## <a name="azure-basic-cluster-setup"></a> Basic Cluster Setup
+
+On the coordinator node:
+
+```bash
+# Setup your test cluster with PostgreSQL 11.5 and Citus master branch
+fab use.postgres:11.5 use.citus:master setup.basic_testing
```
marcocitus: I guess these instructions are not specific to Azure? Should there be a common section on how to use fab?

README.md
```diff
@@ -26,7 +38,213 @@ required for testing citus.
...
+
+## <a name="azure-pgbench"></a> Running PgBench Tests
+
+On the coordinator node:
+
+```bash
+# This will run default pgBench tests with PG=11.5 and Citus Enterprise 9.0 and 8.3 release branches
+# and it will log results to pgbench_results_{timemark}.csv file
+# Yes, that's all :) You can change settings in fabfile/pgbench_confs/pgbench_default.ini
+fab run.pgbench_tests
```
marcocitus: Thought on a future improvement: pgbench is very CPU intensive so running it on the coordinator is likely to skew the benchmark results. It would be nice if we could provision a separate driver node and control the cluster from there.

azure/azuredeploy.parameters.json
```diff
@@ -0,0 +1,27 @@
+{
+    "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentParameters.json#",
+    "contentVersion": "1.0.0.0",
+    "parameters": {
+        "sshPublicKey": {
+            "value": "your public key goes here"
+        },
+        "numberOfInstances": {
+            "value": 4
```
marcocitus: I guess this includes the coordinator. Any chance we can have a "NumberOfWorkers"?

azure/create-cluster.sh
SaitTalhaNisanci: It seems that eastus2 is less expensive, why do we want eastus2?

README.md
SaitTalhaNisanci: I will update this to general test instead of just pgbench, because other tests have timing parameter too. Pgbench is just one of them.

README.md
SaitTalhaNisanci: No they are not. At the end of the readme there is a section that explains the usage of fab. This section existed in AWS so I have copied it for Azure, but this is not necessary. We can remove it if it creates a confusion.

marcocitus: btw, this seems to install postgres multiple times.

azure/azuredeploy.parameters.json
SaitTalhaNisanci: Yeah, that would be more clear.

README.md
marcocitus: Having a common section would be good.

README.md
```diff
@@ -1,23 +1,35 @@
...
+  * [Getting Started](#azure-getting-started)
+    * [Setup steps for each test](#azure-setup-steps)
+    * [Steps to delete a cluster](#azure-delete-cluster)
+    * [Basic Cluster Setup](#azure-basic-cluster-setup)
+    * [Running PgBench Tests](#azure-pgbench)
+    * [Running Scale Tests](#azure-scale)
+    * [Running PgBench Tests Against Citus Cloud](#azure-pgbench-cloud)
+    * [Running TPC-H Tests](#azure-tpch)
+    * [Running TPC-H Tests Against Citus Cloud](#azure-tpch-cloud)
+  * [How it works](#how-it-works)
```
onderkalaci: Maybe rename to `Under the hood` or something like that

README.md
```diff
@@ -26,7 +38,213 @@ required for testing citus.
...
+
+->
+
+pgbench_command: pgbench -c 32 -j 16 -T 5 -P 10 -r
+
+```
+
+</details>
+
+<details>
```
onderkalaci: Maybe `AWS (deprecated)`. I think it is redundant to have `click to expand` note since it is apparent from the icon

README.md
onderkalaci: Agreed, this is too much repetition now as I already noted in my first review as well

README.md
```diff
@@ -26,7 +38,213 @@ required for testing citus.
...
+
+## <a name="azure-getting-started"></a> Getting Started
+
+You can find more information about every step below in other categories. This list of commands show how to get started quickly. Please see other items below to understand details and solve any problems you face.
+
+You should have `az cli` in your local to continue.
+
+You should use `ssh-agent` to add your ssh keys, which will be used for downloading the enterprise repository. Note that your keys are kept only in memory, therefore this is a secure step.
+
+In `azuredeploy.parameters.json` file, you will see the parameters that you can change. For example if you want to change the number of instances, you will need to change the parameter `numberOfInstances`. You can change the type of coordinator and workers separately from the parameters file. Also by default for workers, memory intense vms are used(E type) while for coordinator CPU intense vms are used(D type).
```
onderkalaci: Minor: `azuredeploy.parameters.json` -> `azure/azuredeploy.parameters.json`

or probably add one step below `cd azure` and all other things also would work fine

azure/azuredeploy.parameters.json
```diff
@@ -0,0 +1,27 @@
+{
```
onderkalaci: is it possible to specify the region that we want to spin up the cluster?

README.md
```diff
@@ -26,7 +38,213 @@ required for testing citus.
...
+ssh-add
+
+# Put your public key to sshPublicKey in azuredeploy.parameters.json. You can see your public key with:
+cat ~/.ssh/id_rsa.pub
+
+# in the session that you will use to ssh, set the resource group name
+export RESOURCE_GROUP_NAME=talha_test_resource_group
+
+# Quickly start a cluster of with defaults. This will create a resource group and use it for the cluster.
+./create-cluster.sh
```
onderkalaci: Hmm, I'm currently blocked. I get the following error:

```
azure/create-cluster.sh
{
  "id": "/subscriptions/7fec3109-5b78-4a24-b834-5d47d63e2596/resourceGroups/onder_test_auto",
  "location": "eastus2",
  "managedBy": null,
  "name": "onder_test_auto",
  "properties": {
    "provisioningState": "Succeeded"
  },
  "tags": null,
  "type": "Microsoft.Resources/resourceGroups"
}
[Errno 2] No such file or directory: 'azuredeploy.json'
```

SaitTalhaNisanci: run this in the folder `azure`

azure/create-cluster.sh
onderkalaci: Because we have a lot more capacity there. So, if you want to run a large cluster, you almost always need it (as my experience showed).

I also commented on this to make it configurable if possible as it used to be on ec2.

README.md
SaitTalhaNisanci: this is because `azuredeploy.json` is a relative file path, not absolute file path. We can use an absolute path here later.

azure/azuredeploy.parameters.json
onderkalaci: Similarly, can we specify which subscription to use? Or is that not important?

README.md
```diff
@@ -26,7 +38,213 @@ required for testing citus.
...
+
+# Delete security rule 103 to be able to connect
+./delete-security-rule.sh
```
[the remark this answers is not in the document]
SaitTalhaNisanci: Yeah I can add it, I already have that for the circleci job, so I will take it from there.

README.md
SaitTalhaNisanci: I think it shouldnt matter where we run this script from so I will cd into the 'azure' folder from wherever the script is run with:

```bash
azuredir="${0%/*}"
cd ${azuredir}
```

Now we can run the script from any folder.

azure/azuredeploy.parameters.json
SaitTalhaNisanci: I dont think subscription is important.

There will be an environment variable `AZURE_REGION` if it is set, that value will be used, otherwise the default region will be used.

README.md
```diff
@@ -26,220 +35,182 @@ required for testing citus.
...
+# Add your Github ssh key for enterprise (private) repo
+ssh-add
 
-On your local machine:
-```bash
-# Delete the formation
-# It's a good practice to check deletion status from the cloud formation console
-aws cloudformation delete-stack --stack-name "FormationMetin"
+# Put your public key to sshPublicKey in azuredeploy.parameters.json. You can see your public key with:
+cat ~/.ssh/id_rsa.pub
```
onderkalaci: can we come up with a shell command which sets the `sshPublicKey` in `azuredeploy.paramaters.json` instead of `cat`?

README.md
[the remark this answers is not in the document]
onderkalaci: thanks, this is much easier now

azure/init.sh
```diff
@@ -0,0 +1,123 @@
...
+
+# add the username to sudoers so that sudo command does not prompt password.
+# We do not want the password prompt, because we want to run tests without any user input
+echo '${TARGET_USER}     ALL=(ALL) NOPASSWD:ALL' >>/etc/sudoers
+
+su - ${TARGET_USER} <<'EOSU'
+ # add pg and local binaries to the path
+  echo "export PATH=${HOME}/pg-latest/bin/:${HOME}/.local/bin/:$PATH" >> ${HOME}/.bash_profile
+
+  cd ${HOME} && git clone --branch add-azure-support https://github.com/citusdata/test-automation.git
```
onderkalaci: Not forget to remove `--branch add-azure-support` :)

azure/create-cluster.sh
```diff
@@ -0,0 +1,33 @@
...
+az group deployment create -g ${rg} --template-file azuredeploy.json --parameters @azuredeploy.parameters.json --parameters sshPublicKey="${public_key}" 
+
+connection_string=$(az group deployment show -g ${rg} -n azuredeploy --query properties.outputs.ssh.value)
+
+# remove the quotes 
+connection_string=$(echo ${connection_string} | cut -d "\"" -f 2)
+
+echo "run './connect.sh' to connect to the coordinator, or ALTERNATIVELY:"
+
+echo "run './delete-security-rule.sh' to temporarily disable security rule, and connect with:"
```
SaitTalhaNisanci: I have put this as an alternative way to connect to the coordinator aside, would it be better to remove this completely to simplify the process @onderkalaci?

azure/init.sh
```diff
@@ -0,0 +1,123 @@
...
+python get-pip.py
+
+# install git to clone the repository
+yum install -y git
+
+# this is the username in our instances
+TARGET_USER=pguser
+
+# attach disk and mount it for data
+mkfs.ext4 -F /dev/sdc
```
onderkalaci: How are we sure that it'd be also be `/dev/sdc`? Also, don't we need to add it do `/etc/fstab` to keep it resistant to restarts? Some pointers for [that](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-using-volumes.html)

README.md
```diff
@@ -357,6 +333,94 @@ To get the hostname of the master you can run:
...
+# You can change settings in files under the fabfile/pgbench_confs/ directory
+fab run.pgbench_tests:scale_test.ini
+fab run.pgbench_tests:scale_test_no_index.ini
+fab run.pgbench_tests:scale_test_prepared.ini
+fab run.pgbench_tests:scale_test_reference.ini
+fab run.pgbench_tests:scale_test_foreign.ini
+fab run.pgbench_tests:scale_test_100_columns.ini
+```
+
+## <a name="pgbench-cloud"></a> Running PgBench Tests Against Citus Cloud
```
onderkalaci: I think we should not refer to `CitusCloud` anymore. Can you replace all references to `Hyperscale (Citus)` ?

azure/init.sh
SaitTalhaNisanci:
```
Normally, when we create a new Azure VM, the VM's OS disk name is /dev/sda, the temporary disk name is /dev/sdb.

Then we add a new data disk to this VM, by default, the new data disk name will be /dev/sdc.
```

So in our case it will be always `/dev/sdc`, If we want to attach more than one disk we might need to change this, so I think it is better for now to create an issue covering that(So if someone wants to add multiple disks, they know what to change)

azure/init.sh
SaitTalhaNisanci: Good point, we should do this right before merging

azure/init.sh
SaitTalhaNisanci: #121

onderkalaci: Well, the note is just from a random user and there is even an answer below with `The disk usually becomes /dev/sdc but I've seen it as /dev/sda`.

I'd be in favor of spending some more time and finding a programatic solution instead of hardcoding it. Some pointers are [here](https://docs.microsoft.com/en-us/azure/virtual-machines/linux/attach-disk-portal)

On ec2, I've seen that the disk is attached to other `/dev/sdX`s. And, I don't see any guarantees that `sdc` always the right place.

azure/azuredeploy.parameters.json
```diff
@@ -0,0 +1,27 @@
...
+    "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentParameters.json#",
+    "contentVersion": "1.0.0.0",
+    "parameters": {
+        "sshPublicKey": {
+            "value": "your key will be used from ~/.ssh/id_rsa.pub automatically"
```
onderkalaci: maybe add `if you want to use other key, past it here`

README.md
onderkalaci: This is fixed

README.md
SaitTalhaNisanci: It seems added currently.

azure/init.sh
SaitTalhaNisanci: I have added the logic to find it programatically.

SaitTalhaNisanci:
We were using the attached disk for only `citus-installation` folder, which contains the data for performance tests, I have updated that to be more generic so now we mount the attached disk to the whole `/home/pguser` directory

README.md
```diff
@@ -26,220 +35,190 @@ required for testing citus.
...
 
-# When your cluster is ready, it will prompt you with the connection string, connect to master node
-ssh -A ec2-user@ec2-35-153-66-69.compute-1.amazonaws.com
-```
+# if you want to configure the region
+export REGION=eastus2
 
-On the coordinator node:
+# Quickly start a cluster of with defaults. This will create a resource group and use it for the cluster.
+./create-cluster.sh
```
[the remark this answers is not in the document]
SaitTalhaNisanci: I think you are talking about adding this to the readme right? Because we already cd into `azure` in the script.
I am adding this to readme.

azure/azuredeploy.parameters.json
```diff
@@ -0,0 +1,24 @@
...
+            "value": 2
+        },
+        "diskType": {
+            "value": "Premium_LRS"
+        },
+        "coordinatorVmSize": {
+            "value": "Standard_D4s_v3"
+        },
+        "workerVmSize": {
+            "value": "Standard_E4s_v3"
```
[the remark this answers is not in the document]
SaitTalhaNisanci: Yes, workers are using memory more whereas coordinator is using cpu more. D series are cpu optimized and E series are memory optimized. Therefore choosing E series for workers is better. (This was suggested by Marco).

azure/azuredeploy.json
```diff
@@ -0,0 +1,389 @@
...
+                "Standard_LRS",
+                "Standard_GRS",
+                "Standard_ZRS",
+                "Premium_LRS"
+            ],
+            "metadata": {
+                "description": "Disk type"
+            }
+        },
+        "numberOfWorkers": {
```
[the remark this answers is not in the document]
SaitTalhaNisanci: it seems to be already 3

SaitTalhaNisanci:
@onderkalaci I have created a cluster in aws to see if anything is broken. It seems to be working fine. I havent tested all the functionalities, but I have run pgbench and tpch tests for like an hour, and there wasnt any problem.

azure/azuredeploy.parameters.json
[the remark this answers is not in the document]
onderkalaci: On hyperscale(Citus) all the instances are general purpose and the same type. To be compatible with that, I still have a preference to keep them  all the same. Anyone doing careful performance benchmarks would arrange the types carefully anyway.

marcocitus:
Huh? Hyperscale definitely uses Es_v3 instances for workers.

azure/azuredeploy.parameters.json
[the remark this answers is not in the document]
marcocitus: Please keep E here to be consistent with Hyperscale.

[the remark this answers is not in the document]
onderkalaci: sorry for misguiding here, I remembered wrongly (or while ago all the instances were Ds).

README.md
```diff
@@ -26,220 +35,196 @@ required for testing citus.
...
 
-On your local machine:
-```bash
+# in the session that you will use to ssh, set the resource group name
+export RESOURCE_GROUP_NAME=give_your_name_citus_test_automation_r_g
 
-# Add your EC2 keypair's private key to your agent
-ssh-add path_to_keypair/metin-keypair.pem
+# if you want to configure the region
+export REGION=eastus2
```
onderkalaci: super Minor: I think we should  keep but comment out `export REGION=eastus2`, otherwise everyone would follow this and we wouldn't be able to distribute the load across regions.

SaitTalhaNisanci:
@marcocitus, I have updated the workers to use E type by default.

README.md
SaitTalhaNisanci: makes sense

PR #119, opened 2019-10-21
Fix doing the same steps multiple times in setup

### LINKED ISSUE #55: Setup tasks performs unnecessarily build postgres
I always found myself building postgres when issue any setup.xxx tasks. We should not be building postgres when not needed

### LINKED ISSUE #118: Setup steps are run "number of node" times
Currently setup steps are run "number of node" times, which means that if we have a cluster that has 20 nodes, and if we run the following command on coordinator:

```bash
fab use.postgres:11.5 use.citus:master setup.basic_testing
```

The `basic_testing` step will be run 20 times. And in each step we run all commands with `execute` function. The documentation from execute:

```
    The task will then be executed once per host in its host list, which is
    (again) assembled in the same manner as CLI-specified tasks: drawing from
    :option:`-H`, :ref:`env.hosts <hosts>`, the `~fabric.decorators.hosts` or
    `~fabric.decorators.roles` decorators, and so forth.
```

Which means that in each run of 20 times, we will be sending the commands to all the hosts. This means that we are running the command 19 times unnecessarily. To solve this we should run the basic_testing only from the coordinator, so we should add a role `@roles('master')`  to the task.

This is related to #55.

SaitTalhaNisanci:
Currently setup steps are run "number of node" times, which means that if we have a cluster that has 20 nodes, and if we run the following command on coordinator:

```bash
fab use.postgres:11.5 use.citus:master setup.basic_testing
```

The `basic_testing` step will be run 20 times. And in each step we run all commands with `execute` function. The documentation from execute:

```
    The task will then be executed once per host in its host list, which is
    (again) assembled in the same manner as CLI-specified tasks: drawing from
    :option:`-H`, :ref:`env.hosts <hosts>`, the `~fabric.decorators.hosts` or
    `~fabric.decorators.roles` decorators, and so forth.
```

Which means that in each run of 20 times, we will be sending the commands to all the hosts. This means that we are running the command 19 times unnecessarily. To solve this we should run the basic_testing only from the coordinator, so we should add a role `@roles('master')`  to the task.

Fixes #118.
Fixes #55.

### DIFF
--- a/fabfile/setup.py
+++ b/fabfile/setup.py
@@ -22,6 +22,7 @@ import prefix
 __all__ = ["basic_testing", "tpch", "valgrind", "enterprise"]
 
 @task
+@roles('master')
 def basic_testing():
     'Sets up a no-frills Postgres+Citus cluster'
     execute(prefix.ensure_pg_latest_exists, default=config.CITUS_INSTALLATION)
@@ -30,6 +31,7 @@ def basic_testing():
     execute(add_workers)
 
 @task
+@roles('master')
 def tpch():
     'Just like basic_testing, but also includes some files useful for tpc-h'
     execute(prefix.ensure_pg_latest_exists, default=config.CITUS_INSTALLATION)


fabfile/setup.py
```diff
@@ -30,6 +31,7 @@ def basic_testing():
     execute(add_workers)
 
 @task
+@roles('master')
```
onderkalaci: @SaitTalhaNisanci can you add this to `def enterprise():` as well? I think even for `tpch ` and `valgrind ` we should do similar things, right?

SaitTalhaNisanci: @onderkalaci `enterprise` is not called with `execute`, rather it is directly called, so in that case this is the expected behavior. For `valgrind` there is a sudo command, I wasnt 100% sure if sudo is executed once per host or only in the localhost. For `tpch` we already have it. I am not sure why `enterprise` is not called with `execute` whereas the others are called with `execute`. So there seems to be some inconsistency, I will see if we can simplify this, ideally we can call the functions with execute, and inside functions  we can call everything as a single call not with `execute`, I am guessing this way everything will be run once as we want to. I will also take a look at this.

onderkalaci: I wish we can fix enterprise piece at least, that's annoying :/ @SaitTalhaNisanci

PR #123, opened 2019-10-30
Add periodic performance job

SaitTalhaNisanci:
This PR adds a periodic cron job, which:

- Creates a cluster with the ARM template in Azure with defaults
- Connects to the coordinator
- Runs the tests
- Pushes the results to https://github.com/citusdata/release-test-results

For development purposes the test is done via a regular job, the logic for a periodic job is also added so after getting the other parts approved, it will be changed.


SaitTalhaNisanci:
It seems that there is a time limit for a job in circleci, which is 5 hours.
https://circleci.com/docs/2.0/configuration-reference/#jobs

SaitTalhaNisanci:
Closing this because #128 is merged.

Issue #120, opened 2019-10-22
Tpch tests output causes freeze


SaitTalhaNisanci:
For some of tpch tests, the output is too long therefore it cuts some of the output and puts `--more--` this waits for a user input to continue, something like `q` would make the test continue. However when automating this causes a problem, because we dont want to require any user input to continue running. We can limit the number of rows returned by the tpch queries to solve this problem.

SaitTalhaNisanci:
This is fixed by #128.

PR #128, opened 2019-11-26
Add conditional jobs to run performance tests automatically 

SaitTalhaNisanci:
We have deterministic steps to run pgbench, scale and tpch tests. We still perform many manual steps for running these tests. 
If you want, you can run trigger a job which can run pgbench, scale and tpch tests. What the job does is:

* It creates a cluster with the test resource group name
* It connects to the coordinator
* It runs the corresponding test for the job
* It deletes the cluster.

There is a separate job for each test and you can run any combinations of them. To trigger a job, you should create a branch which has specific prefixes.

If the branch has a prefix `pgbenc`h, then pgbench job will be triggered.
If the branch has a prefix `scale`, then scale job will be triggered.
If the branch has a prefix `tpch`, then tpch job will be triggered.
If the branch has a prefix `all_test`, then all jobs will be triggered.

Each job uses a specific resource group name so that there will be at most 3 resource groups for these jobs. If there is already a resource group, then you should make sure that:

* Someone else is currently not running the same test as you

If not, then you can delete the resource group name from portal, you can find it by search the prefix `citusbot`. Under normal circumstances the resource group will already be deleted at the end of the test
even if it fails.

You can find your test results in https://github.com/citusdata/release-test-results.
By default the tests will be run against `enterprise-master` and the latest released version. If you want to test on a custom branch you should change the config files of relevant tests with your custom branch name in:

```text
postgres_citus_versions: [('12.1', 'your-custom-branch-name-in-enterprise'), ('12.1', 'release-9.1')]
```

Things to be done before merging this PR:
- [x] Test timings should be set back to 300 seconds.
- [x] Branches should be updated as master instead of all_test_try
- [x] Change `all_tests` trigger to `all_performance_tests`
- [x] Put / to the end of all branch names to trigger.


fabfile/pgbench_confs/pgbench_cloud.ini
```diff
@@ -26,10 +26,10 @@ sql_command: SELECT create_distributed_table('pgbench_tellers', 'tid');
...
-pgbench_command: pgbench -c 32 -j 16 -T 300 -P 10 -r
+pgbench_command: pgbench -c 32 -j 16 -T 5 -P 10 -r
 
 [simple-update]
-pgbench_command: pgbench -c 32 -j 16 -T 300 -P 10 -r -b simple-update
+pgbench_command: pgbench -c 32 -j 16 -T 5 -P 10 -r -b simple-update
 
 [select-only]
-pgbench_command: pgbench -c 32 -j 16 -T 300 -P 10 -r -b select-only
+pgbench_command: pgbench -c 32 -j 16 -T 5 -P 10 -r -b select-only
```
marcocitus: 5 seconds?

SaitTalhaNisanci: I have changed it to 5 seconds to test the whole structure fast. I have already added this to the pr description as `things to do before merging`

azure/init.sh
```diff
@@ -40,7 +40,7 @@ su - ${TARGET_USER} <<'EOSU'
  # add pg and local binaries to the path
   echo "export PATH=${HOME}/pg-latest/bin/:${HOME}/.local/bin/:$PATH" >> ${HOME}/.bash_profile
 
-  cd ${HOME} && git clone https://github.com/citusdata/test-automation.git
+  cd ${HOME} && git clone --branch all_test_try https://github.com/citusdata/test-automation.git
```
SaitTalhaNisanci: If we want to change configs, timings we will need to clone that branch here. So we should probably pass the branch here from the circleci job if possible. Otherwise it is kind of tedious to update this and it can be easily forgotten.

azure/delete-resource-group.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
 
 echo "waiting a long time to delete resource group, this might take up to 30 mins depending on your cluster size"
-az group delete -n ${RESOURCE_GROUP_NAME}
+az group delete -n ${RESOURCE_GROUP_NAME} --yes
```
onderkalaci: what is `yes` parameter here?

azure/init.sh
onderkalaci: Yes, that'd be great. If not, we should document to change the branch name here as well.

tpch_2_13_0/distributed_queries/12.sql
```diff
@@ -28,4 +28,5 @@ WHERE
 GROUP BY
     l_shipmode
 ORDER BY
-    l_shipmode;
+    l_shipmode
+LIMIT 4;
```
onderkalaci: why do we change the queries? These are predefined tpch queries, so it might be nicer not to change

README.md
```diff
@@ -35,6 +36,38 @@ required for testing citus.
...
+If you want, you can run trigger a job which can run pgbench, scale and tpch tests. What the job does is:
+
+* It creates a cluster with the test resource group name
+* It connects to the coordinator
+* It runs the corresponding test for the job
+* It deletes the cluster.
+
+There is a separate job for each test and you can run any combinations of them. To trigger a job, you should create a branch which has specific prefixes.
+
+If the branch has a prefix pgbench, then pgbench job will be triggered.
```
onderkalaci: Suggestion: Instead of `all_test` maybe `all_performance_tests` ?

fabfile/pgbench_confs/pgbench_custom.ini
```diff
@@ -1,5 +1,5 @@
 [DEFAULT]
-postgres_citus_versions: [('12.1', 'release-9.1'), ('12.1', 'release-9.0')]
+postgres_citus_versions: [('12.1', 'enterprise-master'), ('12.1', 'release-9.1')]
```
onderkalaci: If it is hard let's skip, but is it possible to fetch the latest release tag and put automatically -- unless a specific value is specified?

For example, can we add a special keyword, like `latest-version` and have
``` 
postgres_citus_versions: [('12.1', 'enterprise-master'), ('12.1', 'latest-version')]
```

That said, if it is going to take more than 1 hour to do that, we can skip

azure/delete-resource-group.sh
SaitTalhaNisanci: deleting a resource group asks for confirmation, if `yes` is passed, it does not. As we cannot have any manual user input we are using `yes`.

tpch_2_13_0/distributed_queries/12.sql
SaitTalhaNisanci: Even when we do manual tpch tests, sometimes the output is long and it shows `more` and it waits for user input to continue. So basically if you run a tpch test manually, you cannot finish the test without pressing enter a few times because of this. I have realized this when automating the tests and therefore limited the results a bit to prevent this.

fabfile/pgbench_confs/pgbench_custom.ini
SaitTalhaNisanci: I guess in fab scripts we can check if it is 'latest-version` if so, we can have a function that returns the latest release we have. I will check if it is easy to do.

SaitTalhaNisanci: I guess this will take more than 1 hour to do but it is a nice thing to have, so we might do it in another PR :)

README.md
SaitTalhaNisanci: Sure we can do that, I will need to do that right before merging so that I dont need to create a separate PR now ( as we have the conversations in this one)
I am adding as to do for right before merging to the description.

azure/init.sh
SaitTalhaNisanci: I have changed it to be done automatically. If it is run in local it will use `master` by default if not it will use the branch name.

tpch_2_13_0/distributed_queries/12.sql
SaitTalhaNisanci: https://github.com/citusdata/test-automation/issues/120

onderkalaci: Then, shall we at least have comments like:
```SQL
....
ORDER BY
    l_shipmode
-- the following LIMIT is not part of the benchmark suite
-- but required to run the tests automatically
-- see https://github.com/citusdata/test-automation/issues/120
LIMIT 5;
```

My goal is to prevent any future users to rely on these queries as being the exact tpch queries.

README.md
```diff
@@ -35,6 +36,38 @@ required for testing citus.
...
+If the branch has a prefix all_test, then all jobs will be triggered.
+
+Each job uses a specific resource group name so that there will be at most 3 resource groups for these jobs. If there is already a resource group, then you should make sure that:
+
+* Someone else is currently not running the same test as you
+
+If not, then you can delete the resource group name from portal, you can find it by search the prefix `citusbot`. Under normal circumstances the resource group will already be deleted at the end of the test
+even if it fails.
+
+You can find your test results in https://github.com/citusdata/release-test-results.
```
onderkalaci: Under which folder of `https://github.com/citusdata/release-test-results`?

README.md
```diff
@@ -35,6 +36,38 @@ required for testing citus.
...
+By default the tests will be run against `enterprise-master` and the latest released version. If you want to test on a custom branch you should change the config files of relevant tests with your custom branch name in:
+
+```text
```
onderkalaci: I think we can improve this readme by improving this  section by:

(a) refer to place in the README where document how to change the cluster size/node count etc
(b)  refer to place in the README where document how to change the pgbench/scale tests params
(c) Very briefly explain what should user in the branch do in order to change the above

PR #129, opened 2019-12-12
Fix rm command call

onurctirtir:
This pr aims to fix `rm` command call made in `setup.py`

### DIFF
--- a/fabfile/setup.py
+++ b/fabfile/setup.py
@@ -70,7 +70,7 @@ def common_setup(build_citus_func):
 
     prefix.check_for_pg_latest()
     # empty it but don't delete the link
-    run('rm -r {}/ || true'.format(config.PG_LATEST))
+    run('rm -r {}/* || true'.format(config.PG_LATEST))
 
     redhat_install_packages()
     build_postgres()


README.md
```diff
@@ -72,7 +72,7 @@ export RESOURCE_GROUP_NAME=give_your_name_citus_test_automation_r_g
 # export REGION=eastus2
 
 # open and modify the instance types/discs as you wish
-less azuredeploy.parameters.json
+vim azuredeploy.parameters.json
```
SaitTalhaNisanci: It is possible that `vim` is not installed in some people whereas `less` will exist. The point here was to show to parameters file and if people want they can change it with their favorite editor.

fabfile/config.py
```diff
@@ -14,7 +14,7 @@
 CITUS_INSTALLATION = os.path.join(HOME_DIR, 'citus-installation')
 
 PG_VERSION = '9.6.1'
-PG_CONFIGURE_FLAGS = ['--with-openssl']
+PG_CONFIGURE_FLAGS = ['--with-openssl', '--with-uuid=e2fs']
```
SaitTalhaNisanci: This will be a problem for any part where we install citus and run tests. I feel like we should try to find some other way about this extension problem. Installing it everywhere does not seem like a good solution to me. Maybe you should start a discussion in team?

fabfile/setup.py
```diff
@@ -70,7 +70,7 @@ def common_setup(build_citus_func):
 
     prefix.check_for_pg_latest()
     # empty it but don't delete the link
-    run('rm -r {}/ || true'.format(config.PG_LATEST))
+    run('rm -r {}/* || true'.format(config.PG_LATEST))
```
SaitTalhaNisanci: Is this to keep the folder there?

README.md
onurctirtir: that makes sense, reverting this commit

fabfile/config.py
onurctirtir: Then I will customize adding this flag for valgrind test. Other than this, I plan to add an optional output to regression outputs

fabfile/setup.py
onurctirtir: As I know, If we want to `# empty it but don't delete the link`, we should call `rm -r softLink/*`. Otherwise, I get error for this `rm` command

SaitTalhaNisanci: What error do you get? We use this function in other tests too, I can check if the error you get exists in other functions as well.

fabfile/config.py
onurctirtir: Removed related commit from this branch and customized it on https://github.com/citusdata/test-automation/pull/130

SaitTalhaNisanci: Yeah I think the ideal solution would be to have optional output in regression tests, I dont know how easy that is.  Because this is adding a dependency to all of our tools etc that use tests.

onurctirtir: Yep, I will add optional output for this soon

fabfile/setup.py
onurctirtir: If we call `rm -r` for a link, we get the following error `rm: cannot remove '<softLinkName>/': Not a directory`

PR #130, opened 2019-12-12
Add valgrind tests

onurctirtir:
This pr aims to 
* migrate valgrind tests from `tools` repository to `test-automation` repository
* automate weekly valgrind tests on circle ci


README.md
```diff
@@ -72,6 +73,7 @@ export RESOURCE_GROUP_NAME=give_your_name_citus_test_automation_r_g
 # export REGION=eastus2
 
 # open and modify the instance types/discs as you wish
+# for valgrind tests, set `numberOfWorkers` to `0` as we do not allocate seperate instances for workers.
```
SaitTalhaNisanci: for valgrind tests, set `numberOfWorkers` to `0` as we do not allocate seperate instances for workers.

->

for valgrind tests, set `numberOfWorkers` to `0` as we only use the coordinator.

README.md
```diff
@@ -430,6 +432,18 @@ On the coordinator node:
 fab run.tpch_automate:tpch_q1.ini,connectionURI='postgres://citus:dwVg70yBfkZ6hO1WXFyq1Q@c.fhhwxh5watzbizj3folblgbnpbu.db.citusdata.com:5432/citus?sslmode\=require'
 ```
 
+## <a name="valgrind"></a> Running Valgrind Tests
+On the coordinator node:
+
+```bash
+# Note that we do not call use.citus(or enterprise) yet as we will do it in run.valgrind task
```
SaitTalhaNisanci: Why dont we use use.citus or enterprise here? We can use the existing logic instead of creating a new one

fabfile/run.py
```diff
@@ -203,3 +202,32 @@ def tpch_queries(query_info, connectionURI, pg_version, citus_version, config_fi
...
+    # validate arguments
+    if len(args) != 2 or args[0] not in ('enterprise', 'citus'):
+        abort('You must provide two arguments, with a command such as "run.valgrind:citus,v6.0.1".\n\
+            First argument can only be "citus" or "enterprise".')
+
+    citus_repo = args[0]
+    citus_version = args[1]
+
+    # install citus and set citus path variable
+    if citus_repo == 'citus':
```
SaitTalhaNisanci: I think we should use the existing logic for deciding whether to download community or enterprise.

fabfile/run.py
```diff
@@ -203,3 +202,32 @@ def tpch_queries(query_info, connectionURI, pg_version, citus_version, config_fi
...
+        repo_path = config.CITUS_REPO
+    else:
+        repo_path = config.ENTERPRISE_REPO
+    
+    setup.valgrind(citus_repo, citus_version)
+
+    with cd(os.path.join(repo_path, config.RELATIVE_REGRESS_PATH)):
+        run('make check-multi-vg VALGRIND_LOG_FILE=logs.txt || true')
```
SaitTalhaNisanci: Can you create a variable in `config.py` and maybe use something like `valgrind_logs.txt` otherwise it is not clear. 

What is the reason for using `` || true`` if the run of valgrind test fails, do we still want to continue? If so why?

fabfile/run.py
```diff
@@ -203,3 +202,32 @@ def tpch_queries(query_info, connectionURI, pg_version, citus_version, config_fi
...
+        
+        # ship output files to result file in order to push to github
+        run('mv logs.txt ' + config.RESULTS_DIRECTORY)
+        run('mv regression.diffs ' + config.RESULTS_DIRECTORY)
```
SaitTalhaNisanci: you can also put regression.diffs in a variable so that we dont have it hardcoded here.

fabfile/setup.py
```diff
@@ -41,19 +42,35 @@ def tpch():
...
-    execute(install_valgrind)
+def valgrind(citus_repo, citus_version):
+
+    # validate and fetch arguments
+
+    if citus_repo not in ('enterprise', 'citus'):
+        abort('You must provide two arguments, with a command such as "run.valgrind:citus,v6.0.1"'
+            'First argument can only be "citus" or "enterprise".')
+
+    config.PG_CONFIGURE_FLAGS.extend(
```
SaitTalhaNisanci: You can create a method in `use.py`:

```python
@task
def valgrind(*args):
    #append the flags for valgrind here
````

File: fabfile/use.py [hunk 7]
```diff
@@ -71,3 +71,8 @@ def asserts(*args):
 def debug_mode(*args):
     '''ps's configure is passed: '--enable-debug --enable-cassert CFLAGS="-ggdb -Og -g3 -fno-omit-frame-pointer"' '''
     config.PG_CONFIGURE_FLAGS.append('--enable-debug --enable-cassert CFLAGS="-ggdb -Og -g3 -fno-omit-frame-pointer"')
+
+
+@task
+def valgrind(*args):
```
SaitTalhaNisanci: We do installation here so this function should be in `setup.py`

README.md
onurctirtir: 1. 
> Did a quick review, one main thing is the usage of community/enterprise. We have functions that set community or enterprise in the config. Why dont we use them?

For some tests, we call directly run command as I see in README. Other than this, I wanted `run.valgrind` task to be aware of the path we are working on in order to move the logs from there. Hence, I parameterized `run.valgrind` so it also specifies repository. As I already parameterized it, using `use.citus/enterprise` may result in conflicts and seems redundant to me. This why I utilize `use.citus/enterprise` in `run.valgrind`.

2. 

> for valgrind tests, set numberOfWorkers to 0 as we do not allocate seperate instances for workers.
> ->
> for valgrind tests, set numberOfWorkers to 0 as we only use the coordinator.

That makes sense

README.md
onurctirtir: For some tests, we call directly run command as I see in README. Other than this, I wanted `run.valgrind` task to be aware of the path we are working on in order to move the logs from there. Hence, I parameterized `run.valgrind` so it also specifies repository. As I already parameterized it, using `use.citus/enterprise` may result in conflicts and seems redundant to me. This why I utilize `use.citus/enterprise` in `run.valgrind`.

fabfile/run.py
onurctirtir: same for https://github.com/citusdata/test-automation/pull/130#discussion_r357102192

fabfile/run.py
onurctirtir: > Can you create a variable in config.py and maybe use something like valgrind_logs.txt otherwise it is not clear.

That makes sense

> What is the reason for using || true if the run of valgrind test fails, do we still want to continue? If so why?

This is just because I migrated the existing linux commands from tools repository :). But it does not makes sense to me as well, I will remove `|| true`

README.md
SaitTalhaNisanci: yeah that seems better :)

README.md
onurctirtir: That makes sense, applied necessary changes

fabfile/run.py
onurctirtir: done

README.md
```diff
@@ -72,6 +73,7 @@ export RESOURCE_GROUP_NAME=give_your_name_citus_test_automation_r_g
 # export REGION=eastus2
 
 # open and modify the instance types/discs as you wish
+# for valgrind tests, set `numberOfWorkers` to `0` as we only need a single machine cluster
```
SaitTalhaNisanci: I think we can be even more explicit here to clarify why we only need a single machine cluster.

Maybe:

`for valgrind tests, set `numberOfWorkers` to `0` as we only need a single machine cluster. This is because we use our regression test structure for valgrind and it creates a local cluster.

azure/script.sh
```diff
@@ -0,0 +1,11 @@
+#eval `ssh-agent -s`
```
SaitTalhaNisanci: Did you push this file by accident?

fabfile/config.py
```diff
@@ -12,6 +12,11 @@
 HOME_DIRECTORY = HOME_DIR
 RESULTS_DIRECTORY = os.path.join(HOME_DIR, 'results')
 CITUS_INSTALLATION = os.path.join(HOME_DIR, 'citus-installation')
+RELATIVE_REGRESS_PATH = 'src/test/regress'
+# valgrind test variables
+VALGRIND_LOGS_FILE = 'logs.txt'
```
SaitTalhaNisanci: `logs.txt` doesnt reveal what this file contains. Maybe rename it to `valgrind_logs.txt`

fabfile/run.py
```diff
@@ -203,3 +202,31 @@ def tpch_queries(query_info, connectionURI, pg_version, citus_version, config_fi
...
+    # install citus and set citus path variable
+    if citus_repo == 'community':
+        repo_path = config.CITUS_REPO
+    elif citus_repo == 'enterprise':
+        repo_path = config.ENTERPRISE_REPO
+    
+    setup.valgrind()
+
+    with cd(os.path.join(repo_path, config.RELATIVE_REGRESS_PATH)):
+        run('make check-multi-vg VALGRIND_LOG_FILE=logs.txt')
```
SaitTalhaNisanci: here you should use the variable `VALGRIND_LOGS_FILE` instead of hard coded `logs.txt`

fabfile/run.py
```diff
@@ -203,3 +202,31 @@ def tpch_queries(query_info, connectionURI, pg_version, citus_version, config_fi
...
+        
+        # ship output files to result file in order to push to github
```
SaitTalhaNisanci: For better readability you can have a method that filters the valgrind logs and a method that moves the outputs. So the structure could be like this:

```python
filter_valgrind_logs()

move_valgrind_outputs()
```

However I have one question, instead of moving the outputs, why dont we directly generate them in the desired output directory(results)?

fabfile/setup.py
```diff
@@ -103,7 +106,7 @@ def redhat_install_packages():
 
     with hide('stdout'):
         sudo('yum install -q -y libxml2-devel libxslt-devel'
-            ' openssl-devel pam-devel readline-devel libcurl-devel git')
+            ' openssl-devel pam-devel readline-devel libcurl-devel git libuuid-devel')
```
SaitTalhaNisanci: is `libuuid-devel` for the extension?

fabfile/use.py
```diff
@@ -19,8 +19,12 @@ def citus(*args):
 
     if len(args) != 1:
         abort('You must provide a single argument, with a command such as "use.citus:v6.0.1"')
+    
     git_ref = args[0]
 
+    # register that we use community
+    config.settings['community-enterprise'] = 'community'
```
SaitTalhaNisanci: Instead of using hard coded strings in different places, can you create variables for these? That way, it is much less error prone. You can put those variables in config. So the motivation is that we should have one place to change them. Currently if you want to change this `community` string you would need to change it at different places.

Also what does `git_ref` hold? Doesnt it already have something like `citus` or `enterprise` ?

azure/script.sh
onurctirtir: exactly, that should not be there. I will remove it

fabfile/config.py
onurctirtir: that makes sense

fabfile/setup.py
onurctirtir: Yes, it is. Now I will parameterize package names to be installed and will add libuuid-devel when the case is valgrind tests

README.md
```diff
@@ -430,6 +433,15 @@ On the coordinator node:
 fab run.tpch_automate:tpch_q1.ini,connectionURI='postgres://citus:dwVg70yBfkZ6hO1WXFyq1Q@c.fhhwxh5watzbizj3folblgbnpbu.db.citusdata.com:5432/citus?sslmode\=require'
 ```
 
+## <a name="valgrind"></a> Running Valgrind Tests
+On the coordinator node:
+
+```bash
+# example usage:
+# Use PostgreSQL 12.1 and run valgrind test on enterprise/enterprise-master
+fab use.valgrind use.postgres:12.1 use.enterprise:enterprise-master run.valgrind
```
SaitTalhaNisanci: I guess you can run `use.valgrind` inside `run.valgrind` because if we know for sure that if someone wants to run valgrind test they will need to run `use.valgrind` as well. So no need to make users put it here.

fabfile/run.py
```diff
@@ -203,3 +202,33 @@ def tpch_queries(query_info, connectionURI, pg_version, citus_version, config_fi
         out_val = run(run_string)
         results_file.write(out_val)
         results_file.write('\n')
+
+def filter_and_put_citus_valgrind_outputs(valgrind_log_path):
+    run('cat {} | grep citus.so | awk \'{ print $1 }\' | uniq  > trace_ids'.format(valgrind_log_path))
```
SaitTalhaNisanci: might be useful to add a comment that explains what this line does

fabfile/use.py
```diff
@@ -39,8 +43,12 @@ def enterprise(*args):
     utils.add_github_to_known_hosts() # make sure ssh doesn't prompt
     if len(args) != 1:
         abort('You must provide a single argument, with a command such as "use.enterprise:v6.0.1"')
+    
     git_ref = args[0]
 
+    # register that we use enterprise
+    config.settings['community-enterprise'] = 'enterprise'
```
SaitTalhaNisanci: I think it might be easier to set the repo path and build function here so that you dont have to write the if else checks in other methods.

```python
config.settings[CITUS_REPO_PATH] = CITUS_ENTERPRISE_REPO
config.settings[CITUS_BUILD_FUNC] = build_enterprise
```

you can do the same for community version:

```python
config.settings[CITUS_REPO_PATH] = CITUS_REPO
config.settings[CITUS_BUILD_FUNC] = build_citus
```

then in valgrind you can directly use `config.settings[CITUS_BUILD_FUNC]` and `config.settings[CITUS_REPO_PATH]`

does that sound good to you?

fabfile/use.py
```diff
@@ -71,3 +79,11 @@ def asserts(*args):
...
     '''ps's configure is passed: '--enable-debug --enable-cassert CFLAGS="-ggdb -Og -g3 -fno-omit-frame-pointer"' '''
     config.PG_CONFIGURE_FLAGS.append('--enable-debug --enable-cassert CFLAGS="-ggdb -Og -g3 -fno-omit-frame-pointer"')
+
+
+@task
+def valgrind(*args):
+    config.PG_CONFIGURE_FLAGS.append('--enable-cassert --enable-debug --with-uuid=e2fs CFLAGS="-ggdb -Og -DUSE_VALGRIND"')
+
+    # set this flag to trigger libuuid-devel installation
+    config.settings['is_valgrind_test'] = True
```
SaitTalhaNisanci: this variable is a leftover I think, it seems that it is not used anywhere.

azure/script.sh
onurctirtir: done

fabfile/run.py
onurctirtir: done

fabfile/run.py
onurctirtir: done

fabfile/setup.py
onurctirtir: yep and removed it as we do not require it anymore

fabfile/use.py
onurctirtir: > Instead of using hard coded strings in different places, can you create variables for these? That way, it is much less error prone. You can put those variables in config. So the motivation is that we should have one place to change them. Currently if you want to change this community string you would need to change it at different places.

done

> Also what does git_ref hold? Doesnt it already have something like citus or enterprise ?

It only keeps name of branch and for some cases, we may not be able to identify branches' repository (like release branches)

README.md
onurctirtir: Upated the comment

fabfile/use.py
onurctirtir: yep, it is removed

fabfile/use.py
onurctirtir: makes sense. done

fabfile/run.py
onurctirtir: done

README.md
onurctirtir: done

onurctirtir:
Now I am running valgrind-test steps to see end-to-end behavior introduced in this pr. But, since  valgrind-tests does also need to push the results to [results repository](https://github.com/citusdata/release-test-results), we need #123 to be merged to observe the full end-to-end behaviour

fabfile/config.py
```diff
@@ -12,6 +12,13 @@
 HOME_DIRECTORY = HOME_DIR
 RESULTS_DIRECTORY = os.path.join(HOME_DIR, 'results')
 CITUS_INSTALLATION = os.path.join(HOME_DIR, 'citus-installation')
+RELATIVE_REGRESS_PATH = 'src/test/regress'
+COMMUNITY_REPO = 'community'
```
SaitTalhaNisanci: This is not used and we dont need this variable there is already a variable called `CITUS_REPO` we can remove this.

fabfile/config.py
```diff
@@ -12,6 +12,13 @@
...
+ENTERPRISE_REPO = 'enterprise'
```
SaitTalhaNisanci: This does not contain the enterprise repo and there is already a variable called `ENTERPRISE_REPO`, which contains the right path for enterprise repo so we should remove this variable too.

onurctirtir: Done

fabfile/config.py
onurctirtir: Done

fabfile/run.py
```diff
@@ -203,3 +202,31 @@ def tpch_queries(query_info, connectionURI, pg_version, citus_version, config_fi
...
+    repo_path = config.settings['repo_path']
+    
+    use.valgrind()
+    setup.valgrind()
+
+    with cd(os.path.join(repo_path, config.RELATIVE_REGRESS_PATH)):
+        run('make check-multi-vg VALGRIND_LOG_FILE=$VALGRIND_LOGS_FILE')
+        
+        # ship regression.diffs to result file in order to push to github
+        run('mv {} {}'.format(config.REGRESSION_DIFFS_FILE, config.RESULTS_DIRECTORY))
```
SaitTalhaNisanci: here we should check if the file exists.

maybe:

```python
if os.path.isfile(config.REGRESSION_DIFFS_FILE):
    run('mv {} {}'.format(config.REGRESSION_DIFFS_FILE, config.RESULTS_DIRECTORY))
```

onurctirtir: we perform this check with the following commit https://github.com/citusdata/test-automation/pull/130/commits/381df184deddcbfd8ac41b8a1bce416beb1f6961

onurctirtir: maybe it would be better to do this check in pythonic way in order not to encounter `fatal error` lines in fab outputs. I will make needed the change

onurctirtir: done in 3b42d6ea24ca75a555ff4a0528d4f2e4c8eda409

fabfile/run.py
```diff
@@ -203,3 +203,57 @@ def tpch_queries(query_info, connectionURI, pg_version, citus_version, config_fi
         out_val = run(run_string)
         results_file.write(out_val)
         results_file.write('\n')
+
+# Filter (possibly) citus-related valgrind test logs, put under results directory
+# If no error logs exist, then simply put valgrind_success file 
+def filter_put_citus_valgrind_outputs(repo_path):
+    # cd to directory that we performed valgrind(regression) tests
+    os.chdir(os.path.join(repo_path, config.RELATIVE_REGRESS_PATH))
```
SaitTalhaNisanci: dont we already use a `cd` wrapper while calling this method, in that case this line can be removed?

fabfile/run.py
```diff
@@ -203,3 +203,57 @@ def tpch_queries(query_info, connectionURI, pg_version, citus_version, config_fi
...
+
+        citus_related_out_exists = True
+
+        if os.path.isfile('trace_ids'):
+            # filter stack traces with stack trace ids that we found above (if any)
+            run('while read line; do grep {} -e $line ; done < trace_ids > {}'.format(
+                config.VALGRIND_LOGS_FILE, 
+                os.path.join(config.RESULTS_DIRECTORY, config.CITUS_RELATED_VALGRIND_LOG_FILE)))
+
+            citus_related_out_exists = True
```
SaitTalhaNisanci: we already set this to true above, so we dont need to set it again to `True`.

fabfile/run.py
```diff
@@ -203,3 +203,57 @@ def tpch_queries(query_info, connectionURI, pg_version, citus_version, config_fi
...
+    if not regression_diffs_exist and not citus_related_out_exists:
+        run('touch {}'.format(os.path.join(config.RESULTS_DIRECTORY, 'valgrind_success')))
+
+@task
+@roles('master')
+def valgrind(*args): 
+    'Runs valgrind tests'
+
+    # set citus path variable
+    repo_path = config.settings['repo_path']
```
SaitTalhaNisanci: you use 'repo_path' as a hard coded value in 3 different places. We should also add this to config as a variable. I would say that whenever you are adding a hard coded string, consider adding it to as a variable in config.

`config.settings[config.REPO_PATH]`.

fabfile/use.py
```diff
@@ -19,8 +21,13 @@ def citus(*args):
 
     if len(args) != 1:
         abort('You must provide a single argument, with a command such as "use.citus:v6.0.1"')
+    
     git_ref = args[0]
 
+    # set community repo specific variables
+    config.settings['repo_path'] = config.CITUS_REPO
+    config.settings['build_citus_func'] = setup.build_citus
```
SaitTalhaNisanci: same as above `config.settings[config.BUILD_CITUS_FUNC]`.

fabfile/run.py
onurctirtir: the `cd` wrapper does not apply to calls made from `os`

fabfile/run.py
onurctirtir: the above should be False, fixing it

SaitTalhaNisanci: should the above be moved outside of the if actually as false? Maybe after `regression_diffs_exist`?

fabfile/use.py
onurctirtir: done

fabfile/run.py
onurctirtir: makes sense, done

fabfile/run.py
onurctirtir: exactly. done

azure/script.sh
```diff
@@ -0,0 +1,15 @@
+export RESOURCE_GROUP_NAME=onurctirtir_test_automation
```
SaitTalhaNisanci: Dont forget to remove this file before merging :)

onurctirtir: yep, it is removed and force-pushed just after

azure/azuredeploy.json
```diff
@@ -40,7 +40,7 @@
         },
         "numberOfWorkers": {
             "defaultValue": 3,
-            "minValue": 1,
+            "minValue": 0,
```
onderkalaci: Do we really need to change this? Aren't we already setting it to `0`  for  valgrind  tests in  another place?

Setting it to zero by default could be a bad user experience

README.md
```diff
@@ -143,9 +144,18 @@ export RESOURCE_GROUP_NAME=give_your_name_citus_test_automation_r_g
 cd azure
 
 # open and modify the instance types/discs as you wish
+#
+# Note: For valgrind tests, setting `numberOfWorkers` to a value is useless as it will be overwritten
```
onderkalaci: I'd rather move this and the below comments to  somewhere else such as `Running Valgrind Tests` as this point is a generic place and, anyone interested in valgrind  test should read the `Running Valgrind Tests` section already

azure/azuredeploy.json
onurctirtir: I initially thought that leaving the `minValue` set to `1` may have problems during some kind of constraint checks, but it did not. Reverted it now

README.md
onurctirtir: done

SaitTalhaNisanci:
There are many more changes after my approval so I am taking my approval back until I do another review.

azure/azuredeploy.json
onurctirtir: Update: if we do not set `minValue` to `0` also, then we get the error below as setting just `numberOfWorkers` to `0` errors out.

`Azure Error: InvalidTemplate
Message: Deployment template validation failed:  'The provided value for the template parameter 'numberOfWorkers' at line '1'`

SaitTalhaNisanci:
What is the status of this PR?

onurctirtir:
> What is the status of this PR?

should be taken care of whenever we are more available I think :/

PR #134, opened 2020-01-05
add hammerdb automation

SaitTalhaNisanci:


hammerdb/build-and-run.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+# fail if trying to reference a variable that is not set.
+set -u
+# exit immediately if a command fails
+set -e
+# echo commands
+set -x
```
JelteF: Should add `set -o pipefail` too, also to other shell scripts.

hammerdb/build-and-run.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
```
JelteF: Please install `shellcheck` and fix/explicitely ignore the warnings it gives. Bash is nice, but very brittle if you don't follow shellcheck.

onderkalaci:
I'm in favor of merging this and adding some documentation on how to use it. This is a great test to have during the release tests.

@SaitTalhaNisanci what do you think?

SaitTalhaNisanci:
Yeah it sounds good, this is not the PR we want to merge though. I will create a PR with separate commits based on the latest branch ( talha_test)

PR #142, opened 2020-01-27
add a script that can generate random prodecures

SaitTalhaNisanci:
This PR adds a script that can create random procedures with their calls. 
It will create 2 files:
- init.sql
- query.sql

`init.sql` contains initializations such as procedure signatures, table definitions, settings etc. Pretty much everything can be configured.

`query.sql` contains procedure calls with random values. The amount of calls is also random which can be configured from the script.

Currently the generated queries are fast path router queries.


generateRandomQueries/README.md
```diff
@@ -0,0 +1,2 @@
+In order to generate a sql file that contains random procedures and calls, use `generate_procedure.py`.
```
onderkalaci: minor: Maybe construct the README such that `procedure` is the first of the random query generation that we implement, and any other new ones can be added more naturally. So, start with a generic intro, and then have a section on `procedure`?

generateRandomQueries/generate_procedure.py
```diff
@@ -0,0 +1,124 @@
...
+
+SETTINGS = [
+    "citus.replication_model TO 'streaming'",
+    "citus.shard_count TO 128"
+]
+SETTING_STRING = "set {};"
+
+MAX_QUERY_PER_PROCEDURE = 300
+MAX_CALL_AMOUNT = 100000
+MIN_TABLE_SIZE = 1000
```
onderkalaci: Minor: table size sounds a little confusing since it might mean disk size etc. So, maybe `MIN_TABLE_ROWS` / `MAX_TABLE_ROWS`?

Also, it might be super useful to have very short descriptions of the variables as comments.  For example, it is not very obvious what `REPEAT_AMOUNT ` mean

generateRandomQueries/README.md
```diff
@@ -0,0 +1,2 @@
+In order to generate a sql file that contains random procedures and calls, use `generate_procedure.py`. 
+This will generate `init.sql` and `query.sql`. `init.sql` contains table definitions, settings, procedures etc. `query.sql` contains producedure calls.
```
onderkalaci: Minor: Maybe mention that this file is used to generate procedures that are pushed down to the workers (uses function call delegation )

PR #143, opened 2020-01-27
remove hard coded value in procedure call

SaitTalhaNisanci:
Generated table have sizes between MIN_TABLE_ROWS and MAX_TABLE_ROWS. However we are calling the procedures with only values 0-100, which is a hard coded value. It makes more sense to call it with values up to MAX_TABLE_ROWS.


SaitTalhaNisanci:
Yeah thanks for finding the bug :)

PR #154, opened 2020-04-16
enhance hammerdb readme

SaitTalhaNisanci:
This PR enhances the hammerdb related parts in the readme. It is to clarify some things a bit more.

### DIFF
--- a/README.md
+++ b/README.md
@@ -418,54 +418,14 @@ pgbench_command: pgbench -c 32 -j 16 -T <test time in seconds> -P 10 -r
 
 ## <a name="running-automated-hammerdb-benchmark"></a>Running Automated Hammerdb Benchmark
 
-Hammerdb tests are run from a driver node. Driver node is in the same virtual network as the cluster.
-You can customize the hammerdb cluster in the `hammerdb` folder using `azuredeploy.parameters.json`.
-
 **You should create a new branch and change the settings in the new branch and push the branch so that
 when the tool clones the repository it can download your branch.**
 
-In order to run hammerdb benchmark:
+Hammerdb tests are run from a driver node. Driver node is in the same virtual network as the cluster.
+You can customize the hammerdb cluster in the `hammerdb` folder using `hammerdb/azuredeploy.parameters.json`.
+Note that this is the configuration for the cluster, which is separate than benchmark configurations(`fabfile/hammerdb_confs/`)
 
-```bash
-eval `ssh-agent -s`
-ssh-add
-export RESOURCE_GROUP_NAME=<your resource group name>
-export GIT_USERNAME=<Your github username>
-export GIT_TOKEN=<Your github token with at least write and read access> # You can create a github token from https://github.com/settings/tokens.
-cd hammerdb
-# YOU SHOULD CREATE A NEW BRANCH AND CHANGE THE SETTINGS IN THE NEW BRANCH
-# AND PUSH THE BRANCH SO THAT WHEN THE TOOL CLONES THE REPOSITORY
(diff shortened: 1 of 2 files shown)


serprex:
I run into this error:

```
waiting a long time to create cluster, this might take up to 30 mins depending on your cluster size
This command is implicitly deprecated because command group 'group deployment' is deprecated and will be removed in a future release. Use 'deployment group' instead.
{'additional_properties': {}, 'code': 'InvalidTemplateDeployment', 'message': "The template deployment 'azuredeploy' is not valid according to the validation procedure. The tracking id is '6ea7af9c-3429-4af2-9526-a11dc75ccd85'. See inner errors for details.", 'target': None, 'details': [<azure.mgmt.resource.resources.v2019_07_01.models._models_py3.ErrorResponse object at 0x7f2a71fc3198>], 'additional_info': None}
```

SaitTalhaNisanci:
@serprex did you change https://github.com/citusdata/test-automation/blob/master/hammerdb/azuredeploy.json by accident maybe?

serprex:
No

SaitTalhaNisanci:
Um, so when I read the error code `InvalidTemplateDeployment` I assumed that there was a problem in the template (such as syntax error) but it seems that it is a core amount limit. So It seems that you cannot create the cluster in westus2 because if you do, you will exceed the quota. 

Can you run `az account list` and see which subscription is your default(`isDefault: true`)? `DS Data Magnet Incubations` We use this, it might be that yours is different for some reason. If so, can you set it as default with `az account set --subscription {uuid-of-correct-subscription}` in this case it should be DS Data Magnet Incubations uuid.

serprex:
My default was on `Visual Studio Enterprise`. Updated

Now it gives

```
Deployment failed. Correlation ID: 16175837-bf8c-4681-8db0-47370276a4c6. The aggregated deployment error is too large. Please list deployment operations to get the deployment details. Please see https://aka.ms/DeployOperations for usage details.
```

serprex:
https://github.com/Azure/azure-cli/issues/13337 is tracking issues which are blocking this

SaitTalhaNisanci:
@serprex I think the issue is resolved on az cli side.

README.md
```diff
@@ -418,24 +418,44 @@ pgbench_command: pgbench -c 32 -j 16 -T <test time in seconds> -P 10 -r
...
 export GIT_USERNAME=<Your github username>
-export GIT_TOKEN=<Your github token with at least write and read access> # You can create a github token from https://github.com/settings/tokens.
+export GIT_TOKEN=<Your github token with write:packages and read:packages permissions> # You can create a github token from https://github.com/settings/tokens.
 cd hammerdb
-# YOU SHOULD CREATE A NEW BRANCH AND CHANGE THE SETTINGS IN THE NEW BRANCH
+# YOU SHOULD CREATE A NEW BRANCH AND CHANGE THE SETTINGS/CONFIGURATIONS IN THE NEW BRANCH
 # AND PUSH THE BRANCH SO THAT WHEN THE TOOL CLONES THE REPOSITORY
 # IT CAN DOWNLOAD YOUR BRANCH.
+vim fabfile/hammerdb_confs/<branch_name>.ini # verify that your custom config file is correct
 ./create-run.sh
```
SaitTalhaNisanci, line 459: As suggested by @serprex, it makes sense to put some command like `echo "successfully started the benchmark"` to the end of ./create-run.sh so that silent failures won't go invisible. For example ssh-keyscan could fail without giving any error message.

SaitTalhaNisanci: Addressed this, merging

PR #156, opened 2020-04-20
change some default hammerdb benchmark settings

SaitTalhaNisanci:
When we want to see the performance change of a branch we should use a
10 node cluster, using 4 node cluster will not give us the results we
want to see.

The default runtime was 200 mins, which is too long. Running for 20-30
mins already gives us the results we want to see, so changing the
default from 200 to 25 makes sense.

IS_TPCC and IS_CH environment variables are added.

### DIFF
--- a/hammerdb/run.tcl
+++ b/hammerdb/run.tcl
@@ -19,7 +19,7 @@ diset tpcc pg_defaultdbase pguser
 diset tpcc pg_storedprocs true
 diset tpcc pg_driver timed
 diset tpcc pg_rampup 3
-diset tpcc pg_duration 200
+diset tpcc pg_duration 60
 diset tpcc pg_timeprofile false
 diset tpcc pg_allwarehouse true
 diset tpcc pg_keyandthink false
--- a/README.md
+++ b/README.md
@@ -490,7 +490,7 @@ After adding the configs `fabfile/hammerdb_confs` could look like:
 Note that running a benchmark with a single config file with a vuuser of 150 and 1M iterations could
 take around 3-4 hours. (the whole process)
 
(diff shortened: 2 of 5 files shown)


hammerdb/run.tcl
```diff
@@ -19,7 +19,7 @@ diset tpcc pg_defaultdbase pguser
 diset tpcc pg_storedprocs true
 diset tpcc pg_driver timed
 diset tpcc pg_rampup 3
-diset tpcc pg_duration 200
+diset tpcc pg_duration 25
```
onderkalaci, line 22: I'd keep it 200 minutes, it could catch memory leaks etc. If anyone  ones a  quick  run, I think  he/she should lower this value. But, the  default that we'd  use in release testing seems better with 200

onderkalaci (approved):
This makes sense to me,  I don't expect anyone to run this unless there is a specific PR or it is release-testing, so 10 nodes sounds more useful.

Still, we can wait @marcocitus to see if this makes sense to him as well, given that he is the one who does this testing most nowadays.

hammerdb/run.tcl
SaitTalhaNisanci: In the hammerdb job on citus engine, we probably would want to have a quicker test, something like 20-30 mins run as in this PR. During release tests it makes more sense to have longer runs. We would hit some error with longer runs on hammerdb because of a bug in fab, which is probably fixed by https://github.com/citusdata/test-automation/pull/164.

We could programmatically change it in hammerdb job on citus engine to be around 20-30 mins and keep the default higher as well.

SaitTalhaNisanci: @marcocitus's suggestion is to standardize on 60 mins I think.

marcocitus (approved):

PR #157, opened 2020-04-22
Add a script that replaces the branch names and pg version

SaitTalhaNisanci:
It is error prone to manually replace branch names in all files. As a solution to that, I have written a script which can be used to update files automatically, an example usage:

```
vim update_files.sh

new_branch1=release-9.3
new_branch2=release-9.2
new_pg_version=12.2

./update_files.sh # this will update all the branches correctly
git diff #make sure that everything is fine
```

### DIFF
--- a/release/README.md
+++ b/release/README.md
@@ -1,0 +1,16 @@
+
+# Usage of update_files.sh
+
+If you will test a custom branch, you can update all the config files with `./update_files.sh`. Say that you 
+will do the release testing for `release-9.2` and `release-9.3`. You can do the following:
+
+```bash
+vim update_files.sh
+
+new_branch1=release-9.3
+new_branch2=release-9.2
+new_pg_version=12.2
+
+./update_files.sh # this will update all the branches correctly
+git diff #make sure that everything is fine
+```
--- a/release/update_files.sh
+++ b/release/update_files.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+
+# fail if trying to reference a variable that is not set.
+set -u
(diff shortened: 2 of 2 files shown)


hanefi:
Can we drop the old version parameters if possible?

If someone wants to update a single file with a different version, `./update_files.sh` will no longer update that file if the old_branch1 is not a match for example.

This is just an idea, if it is harder than I expected, I am totally okay with the current way you implemented this

SaitTalhaNisanci:
The old versions will change, so they won't be fixed. enterprise-master will probably be fixed but release-9.2 etc will change. We might be able to use some regex so that we match postgres_citus_versions: [('12.2', 'release-9.3'), ('12.2', 'release-9.2')] (or postgres_citus_version) and change the relative fields. It should be possible.

Will give an update once I try that.

SaitTalhaNisanci:
@hanefi, I removed the old variables, can you check if this one makes sense to you?

hanefi (approved):
I had a hard time reading the regexes, but they seem to work fine 👍
Well done

SaitTalhaNisanci:
Do you have a suggestion to make regexes more readable?

SaitTalhaNisanci:
One thing we can do is to add a comment to what each regex does.

SaitTalhaNisanci:
@hanefi I have added a comment, do you think it is more readable now?

hanefi:
> @hanefi I have added a comment, do you think it is more readable now?

Way better, thanks for those comments.

PR #159, opened 2020-04-22
Use ssh or https to clone enterprise

### LINKED ISSUE #158: Incorrect way of authentication
We are using git tokens for authentication in hammerdb tool, but other release tests are using ssh keys for authentication. When we are cloning with 'https' ssh keys won't be used and that creates problems in our tpch/scale/pgbench tests currently.

SaitTalhaNisanci:
In test automation we use ssh keys to download enteprise, in hammerdb we
use git tokens to download enterprise. This PR stores that in our config
so that we can use the right method in both of them.

fixes #158.

### DIFF
--- a/fabfile/config.py
+++ b/fabfile/config.py
@@ -1,6 +1,7 @@
 from os.path import expanduser
 import os
 
+IS_SSH_KEYS_USED ='is_ssh_keys_used'
 HOME_DIR = expanduser("~")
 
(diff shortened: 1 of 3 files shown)


SaitTalhaNisanci:
It seems to solve the problem in release tests, we should also verify that it doesn't break hammerdb.

SaitTalhaNisanci:
I verified that this is working with hammerdb, and other release tests.

hanefi (approved):
LGTM

PR #162, opened 2020-04-23
Improve the default values for scale testing

onderkalaci:
As Citus should be able to handle a lot more concurrent queries
as of 9.3, we should increase the defaults.

Max connections is  already `1000` which is more than we need.


SaitTalhaNisanci:
I think we probably want to use 64 clients in scale tests as the memory wasn't good at handling 128 with our default values, but when we increase the default values, we can probably make them 128. There are probably other scale config files where we should change these values I think.

metdos:
Do we want to merge this PR?

SaitTalhaNisanci:
We can possibly first merge https://github.com/citusdata/test-automation/pull/163 then merge this after updating all scale tests with the same default.

PR #164, opened 2020-04-24
Force kill postgres process

### LINKED ISSUE #160: Fabfile: Cannot remove directory
I realized that sometimes we get this error:

```
[10.192.0.8] run: rm -rf /home/pguser/pg-latest/* || true
[10.192.0.8] out: rm: cannot remove �~@~X/home/pguser/pg-latest/data/pg_wal�~@~Y: Directory not empty
[10.192.0.8] out:
```

so it seems that the directory is not being removed even though we use -f option.

This later causes initDb to fail because the directory is not empty.

SaitTalhaNisanci:
It seems that sometimes removing the postgres directory fails, it is
probably because postgres process is still running and therefore we
cannot remove the directory. We send a kill signal to postgres but it
might sometimes fail I guess. Therefore sending a force kill signal
seems a better option.

Fixes #160.

### DIFF
--- a/fabfile/setup.py
+++ b/fabfile/setup.py
@@ -126,7 +126,7 @@ def total_memory_in_gb():
 @parallel
 def common_setup(build_citus_func):
     with hide('stdout'):
-        run('pkill postgres || true')
+        run('pkill -9 postgres || true')
 
     prefix.check_for_pg_latest()
     # empty it but don't delete the link


onderkalaci (approved):
Sometimes `-9` leaves the postmaster.pid on the data dir. However, as we re-init the datadir with `run('rm -rf {}/* || true'.format(config.PG_LATEST))` below, it  looks safe.

PR #165, opened 2020-04-24
not retry if not ssh error

SaitTalhaNisanci:
If we get an error while running the actual scripts, we shouldn't retry them. We should only retry ssh connection errors, in which case 255 will be the returned value.


SaitTalhaNisanci:
This is not going to be a problem in hammerdb side as we use screen there.

onderkalaci (approved):

PR #166, opened 2020-04-24
Support Hammerdb4.0 run

SaitTalhaNisanci:
This branch can be used to run hammerdb 4.0. The reason we need to use an extra variable is that hammerdb4.0 is not yet released and we install it from the package. When it gets released we can remove the `hammerdb_branch` variable and use `hammerdb_version`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -494,6 +494,8 @@ If you want to run only the tpcc benchmark or the analytical queries, you should
 
 You can change the thread count and initial sleep time for analytical queries from `build-and-run.sh` with `CH_THREAD_COUNT` and `RAMPUP_TIME` variables respectively.
 
+If you want to run hammerdb4.0 change `hammerdb_branch` to `hammerdb_4.0_citus` in `create-run.sh`.
+
 ## <a name="basic-cluster-setup"></a> Basic Cluster Setup
 
 On the coordinator node:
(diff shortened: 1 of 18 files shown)


hammerdb/run.tcl
```diff
@@ -23,6 +23,7 @@ diset tpcc pg_duration 200
 diset tpcc pg_timeprofile false
 diset tpcc pg_allwarehouse true
 diset tpcc pg_keyandthink false
+diset tpcc pg_tspace pg_default
```
marcocitus, line 26: this shouldn't be necessary if we include `config/postgresql.xml `

README.md
```diff
@@ -494,6 +494,8 @@ If you want to run only the tpcc benchmark or the analytical queries, you should
 
 You can change the thread count and initial sleep time for analytical queries from `build-and-run.sh` with `CH_THREAD_COUNT` and `RAMPUP_TIME` variables respectively.
 
+If you want to run hammerdb4.0 change `hammerdb_branch` to `hammerdb_4.0_citus` in `create-run.sh`.
```
marcocitus, line 497: This branch name seems a little long to remember. Maybe hammerdb40?

marcocitus (approved):

README.md
SaitTalhaNisanci: For hammerdb3.3 I also updated the branch from `citus` to `hammerdb33` to be consistent.

PR #168, opened 2020-05-04
Add a job to delete resource group for hammerdb automatically

SaitTalhaNisanci:
A job to delete a resource group added. If a branch starts with
`delete_resource_group/` it will delete the resource group by extracting
it from the branch name.

Hammerdb job will push a branch to test automation to trigger this job
so that we don't need to manually delete the resource group after
hammerdb benchmarks are done.

For example, if the resource group name is `citusbot_resource_group_name`,
hammerdb will push the following branch to test automation after
completion: `delete_resource_group/<id>_citusbot_resource_group_name` and
the job will extract the resource group name to delete it.

We can make this deletion optional as well in case we want to ssh to the cluster, but we rarely do. So we can do that optional later as well.

This might create pollution in the repository as we will have many delete_resource_group/ branches. 

One solution is add a nightly job to delete these branches as a cleanup.

Another solution is to remove all delete_resource_group/ branches before pushing to delete_resource/group from the script. Though one downside for this solution is that if there is a concurrent ongoing deletion it will stop that. Though it is not that likely.

I would prefer the first solution as we will need to delete some other branches periodically as well for the hammerdb job on community.

Also it is not a big problem to have many branches in test automation, so we can do it later.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,13 +46,36 @@ jobs:
             ./add-sshkey.sh
             ./citus-bot.sh citusbot_tpch_test_resource_group
           name: install dependencies and run tpch tests
-          no_output_timeout: 10m         
+          no_output_timeout: 10m   
+
+  delete_resource_group:
+    docker:
+      - image: buildpack-deps:buster
+    working_directory: /home/circleci/project  
+    steps:
+      - azure-cli/install
+      - azure-cli/login-with-service-principal
+      - checkout
+      - run:
+          command: |
+            cd ./azure
+            ./delete-resource-group-job.sh
+          name: delete the given resource group
+          no_output_timeout: 10m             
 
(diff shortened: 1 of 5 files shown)


SaitTalhaNisanci:
I have tested this branch, it deletes the resource group as expected after hammerdb is completed.

marcocitus (approved):
That's.. inventive. It would be nicer if we could somehow give the cluster enough access to the Azure API to self-destruct, but seems like it might be tricky.

SaitTalhaNisanci:
>if we could somehow give the cluster enough access to the Azure API to self-destruct, but seems like it might be tricky.

This makes sense, but I couldn't find an easy way for this.  I am not sure if we can delete the cluster from a vm inside the cluster even if we have access to the API.

PR #169, opened 2020-05-07
Remove cmake as we don't need anymore on citus

SaitTalhaNisanci:
https://github.com/citusdata/test-automation/pull/149
https://github.com/citusdata/citus/pull/3626

### DIFF
--- a/azure/init.sh
+++ b/azure/init.sh
@@ -17,17 +17,6 @@ python get-pip.py
 # install git to clone the repository
 yum install -y git screen
 
-# install epel release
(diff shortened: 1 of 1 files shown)


hanefi:
Does this mean that we will not be able to test Citus at a point when we had a cmake dependency?

I guess we haven't released any versions that actually had the cmake dependency. It would be nice to double check this though

SaitTalhaNisanci:
@hanefi  good point, as far as I can see we don't have a version with cmake dependency, @JelteF could verify.

JelteF:
Yes, we never released a version with a cmake dependency.

JelteF (approved):

PR #179, opened 2020-06-24
Always enable debug symbols

JelteF:
This doesn't make execution slower for GCC compiles according to
Postgres docs:

    --enable-debug

        Compiles all programs and libraries with debugging symbols. This
        means that you can run the programs in a debugger to analyze
        problems. This enlarges the size of the installed executables
        considerably, and on non-GCC compilers it usually also disables
        compiler optimization, causing slowdowns. However, having the
        symbols available is extremely helpful for dealing with any
        problems that might arise. Currently, this option is recommended
        for production installations only if you use GCC. But you should
        always have it on if you are doing development work or running a
        beta version.

The `-ggdb` and `-g3` flags simply tell which type of debug symbols we
want.

### DIFF
--- a/README.md
+++ b/README.md
@@ -655,7 +655,7 @@ fab use.enterprise:v7.1.1 setup.enterprise
 fab use.debug_mode use.postgres:10.1 use.citus:v7.1.1 setup.basic_testing
 ```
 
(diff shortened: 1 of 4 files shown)


SaitTalhaNisanci:
Some parts in README could be updated ( mostly parts with debug mode) such as:https://github.com/citusdata/test-automation#-use-tasks. It would be good to mention that by default debug symbols will be enabled.

I am thinking of testing this PR with #180, the change itself makes sense.

PR #180, opened 2020-06-25
Get public key from ssh-agent

JelteF:
That way, if your key is named differently it still works.

### DIFF
--- a/hammerdb/send_pubkey.sh
+++ b/hammerdb/send_pubkey.sh
@@ -6,7 +6,14 @@ set -u
 set -e
 # echo commands
 set -x
+# exit if left size of pipe failed
+set -o pipefail
 
 coordinator_private_ip=$1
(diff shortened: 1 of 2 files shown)


hammerdb/send_pubkey.sh
```diff
@@ -6,7 +6,14 @@ set -u
 set -e
 # echo commands
 set -x
+# exit if left size of pipe failed
+set -o pipefail
 
 coordinator_private_ip=$1
-export pub_key=$(cat ~/.ssh/id_rsa.pub)
+pub_key=$(ssh-add -L | head -n 1)
```
SaitTalhaNisanci, line 13: This script is run on the driver node. Here we are adding the public ip of driver node to the coordinator. If we use ssh agent here, it will be our local key since we will be connecting with ssh forwarding to the coordinator. So it seems correct to use the previous version here, what do you think?

PR #181, opened 2020-06-26
use centos in our vms

SaitTalhaNisanci:
We were using redhat in our vms but it turns out to have some problems.
Currently centos seems to be running smoothly. So switching to centos
makes sense, and it doesn't require many changes. We can go back to
redhat or even make it configurable easily.

In another PR the versions will be increased to the latest 7.X.

### DIFF
--- a/azure/init.sh
+++ b/azure/init.sh
@@ -6,13 +6,13 @@ set -u
 set -e
 # fail in a pipeline if any of the commands fails
 set -o pipefail
+# echo commands
+set -x
 
 # in redhat we need to enable default port for postgres
-firewall-cmd --add-port=5432/tcp
-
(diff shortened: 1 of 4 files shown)


azure/init.sh
```diff
@@ -6,13 +6,13 @@ set -u
 set -e
 # fail in a pipeline if any of the commands fails
 set -o pipefail
+# echo commands
+set -x
 
 # in redhat we need to enable default port for postgres
-firewall-cmd --add-port=5432/tcp
-
-# install azure client rpm to have correct certificates
```
SaitTalhaNisanci, line 13: https://github.com/citusdata/test-automation/pull/173 

We don't need this weird thing anymore.

JelteF (approved):

PR #197, opened 2020-11-02
Enable repartition joins in tpch and update executor

SaitTalhaNisanci:
We should revisit our tests one by one and do necessary changes because some of them are outdated and some tests might not have any value. But it is a bigger thing

### DIFF
--- a/fabfile/tpch_confs/tpch_default.ini
+++ b/fabfile/tpch_confs/tpch_default.ini
@@ -1,7 +1,7 @@
 [test]
 use_enterprise: on
 postgres_citus_versions: [('12.1', 'enterprise-master'), ('12.1', 'release-9.2')]
-tpch_tasks_executor_types: [('1.sql', 'real-time'), ('3.sql', 'task-tracker'), ('5.sql', 'task-tracker'), ('6.sql', 'real-time')
-                            ,('7.sql', 'task-tracker'), ('8.sql', 'task-tracker'), ('9.sql', 'task-tracker'), ('10.sql', 'task-tracker')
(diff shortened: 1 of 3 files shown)


fabfile/tpch_confs/tpch_default.ini
```diff
@@ -1,7 +1,7 @@
 [test]
 use_enterprise: on
 postgres_citus_versions: [('12.1', 'enterprise-master'), ('12.1', 'release-9.2')]
-tpch_tasks_executor_types: [('1.sql', 'real-time'), ('3.sql', 'task-tracker'), ('5.sql', 'task-tracker'), ('6.sql', 'real-time')
-                            ,('7.sql', 'task-tracker'), ('8.sql', 'task-tracker'), ('9.sql', 'task-tracker'), ('10.sql', 'task-tracker')
-                            , ('12.sql', 'real-time'), ('14.sql', 'task-tracker'), ('19.sql', 'task-tracker')]
+tpch_tasks_executor_types: [('1.sql', 'adaptive'), ('3.sql', 'adaptive'), ('5.sql', 'adaptive'), ('6.sql', 'adaptive')
```
onderkalaci, line 4: I guess this syntax doesn't make sense anymore, but to keep the changes small, I'm ok to keep it as it

SaitTalhaNisanci: I think we should go over all of these things and do the relevant changes but it will probably take a while maybe 2-3 days

SaitTalhaNisanci: here I aimed to fix it with minimum changes as you said. With this I was able to get results

onderkalaci (approved):

PR #202, opened 2021-01-29
Cache git credentials in init.sh

onurctirtir:


onurctirtir:
With this pr, we had "unbound variable" errors in init.sh, so reverted cdb15e80f2a6351465ea558c2f2e1da7c21e1614 on master and included it in a new pr #203

PR #208, opened 2021-02-03
Use make install-all when installing citus if available

onurctirtir:
This is mainly to make valgrind testing happy
Tested with valgrind testing, works fine

### DIFF
--- a/fabfile/setup.py
+++ b/fabfile/setup.py
@@ -241,8 +241,8 @@ def build_enterprise():
 def install_citus(core_count):
     with hide('stdout'):
         # fall back to "make install" if "make install-all" is not available
-        run('make -s -j{} install-all || make -s -j{} install'.
-            format(core_count, core_count))
+        run('make -s -j{core_count} install-all || make -s -j{core_count} install'.\
+            format(core_count=core_count))
 
 def create_database():
     pg_latest = config.PG_LATEST
--- a/fabfile/setup.py
+++ b/fabfile/setup.py
@@ -217,7 +217,7 @@ def build_citus():
             core_count = run('cat /proc/cpuinfo | grep "core id" | wc -l')
 
         with hide('stdout'):
-            run('make -s -j{} install'.format(core_count))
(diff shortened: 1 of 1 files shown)


onurctirtir:
Tested with valgrind testing, works fine

SaitTalhaNisanci:
What if we are installing an old version which doesn't have this install-all target?

onurctirtir:
> What if we are installing an old version which doesn't have this install-all target?

maybe `make install-all || make install` ?

SaitTalhaNisanci:
>maybe make install-all || make install ?

Makes sense.

fabfile/setup.py
```diff
@@ -237,8 +236,13 @@ def build_enterprise():
...
         core_count = run('cat /proc/cpuinfo | grep "core id" | wc -l')
 
-        with hide('stdout'):
-            run('make -s -j{} install-all'.format(core_count))
+        install_citus(core_count)
+
+def install_citus(core_count):
+    with hide('stdout'):
+        # fall back to "make install" if "make install-all" is not available
+        run('make -s -j{} install-all || make -s -j{} install'.
```
SaitTalhaNisanci, line 244: Minor:
```python
make -s -j{core_count} install-all || make -s -j{core_count} install'.format(core_count=core_count)
```

SaitTalhaNisanci (approved):

fabfile/setup.py
```diff
@@ -237,8 +236,13 @@ def build_enterprise():
...
 
-        with hide('stdout'):
-            run('make -s -j{} install'.format(core_count))
+        install_citus(core_count)
+
+def install_citus(core_count):
+    with hide('stdout'):
+        # fall back to "make install" if "make install-all" is not available
+        run('make -s -j{} install-all || make -s -j{} install'.
+            format(core_count, core_count))
```
onurctirtir, line 245:
```suggestion
        run('make -s -j{core_count} install-all || make -s -j{core_count} install'.\
            format(core_count=core_count))
```

PR #209, opened 2021-02-05
Use upstream hammerdb

SaitTalhaNisanci:
When we use a custom hammerdb branch with citus related changes, the
branch easily get outdated. It makes more sense to apply only the
relevant patch to the upstream.

### DIFF
--- a/hammerdb/driver-init.sh
+++ b/hammerdb/driver-init.sh
@@ -19,7 +19,7 @@ python get-pip.py
 rpm -Uvh https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
 # install git to clone the repository
 # install screen so that we can run commands in a detached session
-yum install -y git screen tmux htop
+yum install -y git screen tmux htop patch
 
 # this is the username in our instances
 TARGET_USER=pguser
--- a/hammerdb/setup.sh
+++ b/hammerdb/setup.sh
@@ -16,7 +16,7 @@ hammerdb_version=$(cat ~/HAMMERDB_VERSION)
 hammerdb_dir="${HOME}"/HammerDB-"${hammerdb_version}"
 
 cd "${HOME}"
-git clone https://github.com/citusdata/ch-benchmark.git
+git clone -b test-automation https://github.com/citusdata/ch-benchmark.git
 cd ch-benchmark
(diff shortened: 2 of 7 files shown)


hammerdb/driver-init.sh
```diff
@@ -19,7 +19,7 @@ python get-pip.py
 rpm -Uvh https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
 # install git to clone the repository
 # install screen so that we can run commands in a detached session
-yum install -y git screen tmux htop
+yum install -y git screen tmux htop patch
```
SaitTalhaNisanci, line 22: this is needed since we are patching the citus related changes

hammerdb/setup.sh
```diff
@@ -9,27 +9,17 @@ set -x
...
-# here we use our fork, because it distributed tables at the beginning, which speeds up the process
-# since we can create indexes in parallel etc.
-git clone --branch "${hammerdb_branch}" https://github.com/citusdata/HammerDB.git
-mv HammerDB/src/postgresql/pgoltp.tcl "${hammerdb_dir}"/src/postgresql/pgoltp.tcl
-mv HammerDB/config/postgresql.xml "${hammerdb_dir}"/config/postgresql.xml
-
-# cd ${hammerdb_dir}/src/postgresql
-# comment out create database and user as citus cannot do that
-# sed -i 's/CreateUserDatabase $lda $db $superuser $user $password/#CreateUserDatabase $lda $db $superuser $user $password/g' pgoltp.tcl
+git clone https://github.com/citusdata/ch-benchmark.git
```
SaitTalhaNisanci, line 19: this would mean we rely on ch-benchmark, which is probably okay and we have a single place for the citus related changes

marcocitus: I could see us changing the UX a bit if we want to encourage customers to run it. Should we make a separate branch?

SaitTalhaNisanci: separate branch on the ch-benchmark repo you mean?

SaitTalhaNisanci: ping

marcocitus: yes

SaitTalhaNisanci: created `test-automation` branch

PR #210, opened 2021-02-09
Add use_enterprise config to update_files script

naisila:
Add use_enterprise config to update_files script

### DIFF
--- a/release/update_files.sh
+++ b/release/update_files.sh
@@ -29,5 +29,5 @@ find . -type f -name "*.ini" |
 # replace all .ini config files (except master.ini which requires use_enterprise to be off):
 # use_enterprise: <use_enterprise>
 # use_enterprise: <new_use_enterprise>
-find . -type f -name "*.ini" ! -name "master.ini" | 
+find . -type f -name "*.ini" | 
     xargs sed -i "s@use_enterprise: [^,;]\+@use_enterprise: ${new_use_enterprise}@g"
--- a/release/update_files.sh
+++ b/release/update_files.sh
@@ -11,6 +11,7 @@ set -x
 new_branch1=release-9.3
 new_branch2=release-9.2
 new_pg_version=12.2
+new_use_enterprise=on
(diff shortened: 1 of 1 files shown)


release/update_files.sh
```diff
@@ -25,3 +26,8 @@ find . -type f -name "*.ini" |
 # postgres_citus_versions: [('<new_pg_version>', '<new_branch1>')]
 find . -type f -name "*.ini" | 
     xargs sed -i "s@postgres_citus_versions: \[('[^,;]\+', '[^,;]\+')\]@postgres_citus_versions: \[('${new_pg_version}', '${new_branch1}')\]@g"
+# replace all .ini config files (except master.ini which requires use_enterprise to be off):
+# use_enterprise: <use_enterprise>
+# use_enterprise: <new_use_enterprise>
+find . -type f -name "*.ini" ! -name "master.ini" |
```
SaitTalhaNisanci, line 32: I think we can keep `master.ini` as well because we will not use this for hammerdb and not excluding anything seems simpler while doing the job

SaitTalhaNisanci:
You don't need to add "DESCRIPTION" to the description. In citus, we do that because there is a script that finds lines with that prefix. Here we don't have such a prefix.

release/update_files.sh
```diff
@@ -25,3 +26,8 @@ find . -type f -name "*.ini" |
 # postgres_citus_versions: [('<new_pg_version>', '<new_branch1>')]
 find . -type f -name "*.ini" | 
     xargs sed -i "s@postgres_citus_versions: \[('[^,;]\+', '[^,;]\+')\]@postgres_citus_versions: \[('${new_pg_version}', '${new_branch1}')\]@g"
+# replace all .ini config files (except master.ini which requires use_enterprise to be off):
```
SaitTalhaNisanci, line 29: Comment outdated

SaitTalhaNisanci (approved):

PR #212, opened 2021-02-11
Expose multiple ports on ssh

### LINKED ISSUE #211: Remove/fix security rule deletion
https://github.com/citusdata/test-automation/blob/master/hammerdb/connect-driver.sh deletes a security rule, but that's against Azure policy.

SaitTalhaNisanci:
Expose multiple ports on ssh so that we don't hit security group rules when trying to connect via the default port, 22.

The custom port we now use for ssh is 3456.

Alternatively I have tried to use a loadbalancer to have port forwarding but it seems that in that case we again hit rule 103 because only internal requests from loadbalancer to vms are allowed and public requests are not.

Fixes #211.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,8 @@ The data will be stored on the attached disk, size of which can be configured in
 
 If you dont specify the region, a random region among `eastus`, `west us 2` and `south central us` will be chosen. This is to use resources uniformly from different regions.
 
+**Port 3456 is used for ssh, you can connect to any node via Port 3456, if you don't use this node, you will hit the security rules.**
+
 ## <a name="azure-setup-steps"></a> Setup Steps For Each Test
 
 You will need to follow these steps to create a cluster and connect to it, on your local machine:
@@ -98,14 +100,6 @@ less azuredeploy.parameters.json
 
 # connect to the coordinator
 ./connect.sh
-
-# ALTERNATIVATELY instead of ./connect.sh you can do the following
-
-# Delete security rule 103 to be able to connect
-./delete-security-rule.sh
-
-# When your cluster is ready, it will prompt you with the connection string, connect to coordinator node
-ssh -A pguser@<public ip of coordinator>
 ```
 
 ## <a name="azure-delete-cluster"></a> Steps to delete a cluster
@@ -140,9 +134,9 @@ The first virtual machine with index 0 is treated as a coordinator. When all the
 
 The initialization script also finds the private ip addresses of workers and puts them to the coordinator. The way this is done is with a storage account resource. This storage account resource is created within the template itself and all the vms upload their private ip addresses to the storage. After all are uploaded the coordinator downloads all the private ip addresses from the storage account and puts it to `worker-instances` file, which is then used when creating a citus cluster.
 
-We have a special security group which blocks ssh traffic. The rule's priority is 103 and 100, 101, 102 are also taken by this security group. The workaround to connect to the coordinator is to remove the rule 103, and connect right after it. The rule comes back in 2-3 mins, so you should be fast here. There is a script called `delete-security-rule.sh`, which deletes that rule for you.
+We have a special security group which blocks ssh traffic. The rule's priority is 103 and 100, 101, 102 are also taken by this security group.
 
-You can use `connect.sh` which will delete the security rule and connect to the coordinator for you.
+You can use `connect.sh` which will connect to the coordinator for you on a custom ssh port (at the time of writing 3456).
 
 Before starting the process you should set the environment variable `RESOURCE_GROUP_NAME`, which is used in all scripts.
 
@@ -170,13 +164,6 @@ then you should run:
 
 ```bash
 ./connect.sh
-```
-
-or
-
-```bash
-./delete-security-rule.sh
-ssh -A pguser@<public ip of coordinator>
 ```
 
 After you are done with testing, you can delete the resource group with:
--- a/azure/azuredeploy.json
+++ b/azure/azuredeploy.json
@@ -401,7 +401,7 @@
                 "settings": {
                     "commandToExecute": "[concat('sh init.sh ', copyIndex(), ' ', variables('numberOfInstances'), ' ', variables('storageAcctName'), ' ', listKeys(variables('storageAcctName'),'2017-10-01').keys[0].value, ' ', parameters('branchName'))]",
                     "fileUris": [
-                        "https://raw.githubusercontent.com/citusdata/test-automation/multiple_ports_on_ssh/azure/init.sh"
+                        "https://raw.githubusercontent.com/citusdata/test-automation/master/azure/init.sh"
                     ]
                 }
             },
--- a/azure/azuredeploy.json
+++ b/azure/azuredeploy.json
@@ -386,7 +386,7 @@
                 "settings": {
                     "commandToExecute": "[concat('sh init.sh ', copyIndex(), ' ', variables('numberOfInstances'), ' ', variables('storageAcctName'), ' ', listKeys(variables('storageAcctName'),'2017-10-01').keys[0].value, ' ', parameters('branchName'))]",
                     "fileUris": [
-                        "https://raw.githubusercontent.com/citusdata/test-automation/master/azure/init.sh"
+                        "https://raw.githubusercontent.com/citusdata/test-automation/multiple_ports_on_ssh/azure/init.sh"
                     ]
                 }
             },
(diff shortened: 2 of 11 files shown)


azure/azuredeploy.json
```diff
@@ -386,7 +401,7 @@
                 "settings": {
                     "commandToExecute": "[concat('sh init.sh ', copyIndex(), ' ', variables('numberOfInstances'), ' ', variables('storageAcctName'), ' ', listKeys(variables('storageAcctName'),'2017-10-01').keys[0].value, ' ', parameters('branchName'))]",
                     "fileUris": [
-                        "https://raw.githubusercontent.com/citusdata/test-automation/master/azure/init.sh"
+                        "https://raw.githubusercontent.com/citusdata/test-automation/multiple_ports_on_ssh/azure/init.sh"
```
SaitTalhaNisanci, line 404: Will update this to master before merging

README.md
```diff
@@ -98,14 +100,6 @@ less azuredeploy.parameters.json
...
 # connect to the coordinator
 ./connect.sh
-
-# ALTERNATIVATELY instead of ./connect.sh you can do the following
-
-# Delete security rule 103 to be able to connect
-./delete-security-rule.sh
-
-# When your cluster is ready, it will prompt you with the connection string, connect to coordinator node
-ssh -A pguser@<public ip of coordinator>
```
onurctirtir, line 108: instead of removing this line, maybe we should provide port 3456 too

README.md
```diff
@@ -172,13 +166,6 @@ then you should run:
 ./connect.sh
 ```
 
-or
-
-```bash
-./delete-security-rule.sh
-ssh -A pguser@<public ip of coordinator>
```
onurctirtir, line 179: same

azure/create-cluster.sh
```diff
@@ -62,5 +62,4 @@ connection_string=$(echo ${connection_string} | cut -d "\"" -f 2)
 
 echo "run './connect.sh' to connect to the coordinator, or ALTERNATIVELY:"
 
-echo "run './delete-security-rule.sh' to temporarily disable security rule, and connect with:"
 echo ${connection_string}
```
onurctirtir, line 65:
```suggestion
echo "and connect with: ${connection_string}"
```

azure/connect.sh
```diff
@@ -15,8 +15,9 @@ azuredir="${0%/*}"
 cd ${azuredir}
 
 public_ip=$(az group deployment show -g ${RESOURCE_GROUP_NAME} -n azuredeploy --query properties.outputs.publicIP.value)
+ssh_port=$(az group deployment show -g ${RESOURCE_GROUP_NAME} -n azuredeploy --query properties.outputs.customSshPort.value)
+
 # remove the quotes 
 public_ip=$(echo ${public_ip} | cut -d "\"" -f 2)
-sh ./delete-security-rule.sh
-
-ssh -o "StrictHostKeyChecking no" -A pguser@${public_ip}
```
onurctirtir, line 22: minor: maybe one more newline as we did previously

azure/init.sh
```diff
@@ -43,6 +45,17 @@ rsync -aXS /tmp/home_copy/. /home/${TARGET_USER}/.
 # We do not want the password prompt, because we want to run tests without any user input
 echo '${TARGET_USER}     ALL=(ALL) NOPASSWD:ALL' >>/etc/sudoers
 
+# we will use port 3456 to not hit security rule 103
+echo 'Port 3456' >> /etc/ssh/sshd_config
+echo 'Port 22' >> /etc/ssh/sshd_config
```
onurctirtir, line 50: why putting port 22 ?

hammerdb/azuredeploy.json
```diff
@@ -248,6 +249,20 @@
                             "priority": 101,
                             "direction": "Inbound"
                         }
+                    },
+                    {
+                        "name": "customSshPort",
```
onurctirtir, line 254: this might deserve a comment like:

add security rule to enable custom ssh port

hammerdb/create-run.sh
```diff
@@ -23,9 +23,7 @@ ssh_execute() {
    n=0
    until [ $n -ge 10 ]
    do
-      # delete the security before each try
-      sh "${topdir}"/azure/delete-security-rule.sh
-      ssh -o "StrictHostKeyChecking no" -A pguser@"${ip}" "source ~/.bash_profile;${command}" && break
+      ssh -o "StrictHostKeyChecking no" -A -p 3456 pguser@"${ip}" "source ~/.bash_profile;${command}" && break
```
onurctirtir, line 26: can we somehow deduce custom ssh port instead of hardcoding it (if it wouldn't require any additional efforts)?

hammerdb/driver-init.sh
```diff
@@ -40,6 +41,18 @@ rsync -aXS /tmp/home_copy/. /home/"${TARGET_USER}"/.
 # We do not want the password prompt, because we want to run tests without any user input
 echo '${TARGET_USER}     ALL=(ALL) NOPASSWD:ALL' >>/etc/sudoers
 
+# we will use port 3456 to not hit security rule 103
+echo 'Port 3456' >> /etc/ssh/sshd_config
+echo 'Port 22' >> /etc/ssh/sshd_config
```
onurctirtir, line 46: again, why putting 22 ?

azure/init.sh
onurctirtir: I guess it's because now we override ssh ports for custom port and so the default port would go away if we don't add it explicitly

azure/init.sh
```diff
@@ -43,6 +45,17 @@ rsync -aXS /tmp/home_copy/. /home/${TARGET_USER}/.
...
+
+# necessary for semanage, VMs have secure linux
+yum install -y policycoreutils-python
+# we need to enable the new port from semanage
```
onurctirtir, line 54: [ let me do some research for what is semanage :) ]

README.md
SaitTalhaNisanci: Do you think it is valuable? `./connect.sh` seems enough. There wasn't any connect script before and there was only this alternative in the past, so now using `connect.sh` is what we do

azure/init.sh
SaitTalhaNisanci: I kept the default port because Azure has rules based on rule 22 and it tries to connect on 22. So I thought it could also cause some problems and possibly creation of ICM if rule 22 is not working for inner communication in azure.

README.md
onurctirtir: true, it doesn't seem valuable

azure/init.sh
SaitTalhaNisanci: we have secure linux and in secure linux, we need to do some extra commands to enable some security related stuff. semanage helps with adding the port in this case

hammerdb/azuredeploy.json
SaitTalhaNisanci: You mean a separate comment other than the current description 2 lines below or update the description below?

azure/create-cluster.sh
onderkalaci: I prefer `echo ${connection_string}` as it makes it easy to copy & paste.

> ALTERNATIVELY:

Maybe `ALTERNATIVELY RUN THE FOLLOWING:`

hammerdb/azuredeploy.json
```diff
@@ -248,6 +249,20 @@
...
+                        "properties": {
+                            "description": "custom ssh port",
+                            "protocol": "Tcp",
+                            "sourcePortRange": "*",
+                            "destinationPortRange": "[variables('customSshPort')]",
+                            "sourceAddressPrefix": "*",
+                            "destinationAddressPrefix": "*",
+                            "access": "Allow",
+                            "priority": 102,
```
onderkalaci, line 263: what happens if there is already a rule with `"priority": 102,`? Or, could that happen?

SaitTalhaNisanci: there is already a rule with 102 :)

It gets appended after all the existing rules. So 100, 101, 102, and 103 will be reserved. and these rules will have priority of 104, 105...

hammerdb/azuredeploy.json
onurctirtir: > update the description below?

makes more sense, maybe:

"description": "allow accessing to custom ssh port from all ip's",

onderkalaci:
btw, I tested this and works as it used to be, thanks

onurctirtir (approved):
let's not forget to revert those before merging

azure/create-cluster.sh
SaitTalhaNisanci: @onurctirtir FYI I changed this as `ALTERNATIVELY RUN THE FOLLOWING:` as it can indeed easier for copy & paste, I thought it was okay for you since you approved, let me know if it is not

SaitTalhaNisanci:
Thanks for the quick review @onurctirtir.

PR #214, opened 2021-02-15
Not use ssh-keyscan in local as we already don't check key

SaitTalhaNisanci:
We were using ssh-keyscan but it is no longer needed as we already pass
"StrictHostKeyChecking no" option. This would sometimes fail silently
hence causing confusing for developers.

### DIFF
--- a/hammerdb/create-run.sh
+++ b/hammerdb/create-run.sh
@@ -72,9 +72,6 @@ driver_private_ip=$(az group deployment show -g "${cluster_rg}" -n azuredeploy -
 driver_private_ip=$(echo "${driver_private_ip}" | cut -d "\"" -f 2)
 echo "${driver_private_ip}"
 
-# just to be sure, the permission of known_hosts should be 600, otherwise it is ignored.
-chmod 600 ~/.ssh/known_hosts
-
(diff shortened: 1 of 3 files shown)


hammerdb/create-run.sh
```diff
@@ -72,9 +72,6 @@ driver_private_ip=$(az group deployment show -g "${cluster_rg}" -n azuredeploy -
 driver_private_ip=$(echo "${driver_private_ip}" | cut -d "\"" -f 2)
 echo "${driver_private_ip}"
 
-# add the cluster and driver ip to our known host so that they don't prompt any verification.
-ssh-keyscan -H "${cluster_ip}" >> ~/.ssh/known_hosts
-ssh-keyscan -H "${driver_ip}" >> ~/.ssh/known_hosts
 # just to be sure, the permission of known_hosts should be 600, otherwise it is ignored.
 chmod 600 ~/.ssh/known_hosts
```
marcocitus, line 76: does this chmod still make sense?

SaitTalhaNisanci: it doesn't have any harm but it is not necessary anymore I think, so I removed it.

marcocitus (approved):

PR #217, opened 2021-02-20
Only allow our local ip for the custom ssh port

SaitTalhaNisanci:
Azure seems to be not happy about exposing a port to the internet hence
we switch to exposing it to only our public ip. Also in the utility
scripts we update the custom port source ip as our current local ip as
our ip might change.


onurctirtir:
unfortunately local ip would change after a time unless the person that does testing buys static ip service from isp

SaitTalhaNisanci:
>unfortunately local ip would change after a time unless the person that does testing buys static ip service from isp

That is why we add the local (new) ip again when using the scripts such as `./connect-driver` or `.connect.sh`

onurctirtir:
> > unfortunately local ip would change after a time unless the person that does testing buys static ip service from isp
> 
> That is why we add the local (new) ip again when using the scripts such as `./connect-driver` or `.connect.sh`

I didn't realize that, thanks

SaitTalhaNisanci:
For example for finalizing the valgrind, it will already be a different machine. So the ip will be different so we have `./add-local-ip.sh
` which updates the security rule with the current ip, but also the old one is gone. I guess that is okay.

onurctirtir (approved):

PR #233, opened 2021-09-06
Handle pgbench output difference with PG14

SaitTalhaNisanci:


SaitTalhaNisanci:
I tested this, it seems to be working fine, but I suggest that we remove TPC including connection establishment field for older versions as well so that we can print the same thing for all PG versions, any thoughts @onderkalaci ?

PR #235, opened 2021-09-07
Add two more valgrind tasks: multi-1 and columnar 

onurctirtir:


fabfile/run.py
```diff
@@ -13,7 +13,7 @@
 import ConfigParser
 import time
 
-__all__ = ['jdbc', 'regression', 'pgbench_tests', 'tpch_automate', 'valgrind', 'valgrind_filter_put_results']
+__all__ = ['jdbc', 'regression', 'pgbench_tests', 'tpch_automate', 'valgrind_multi', 'valgrind_multi_1', 'valgrind_columnar']
```
onurctirtir, line 16: could instead make use of the `run.valgrind:<option>` syntax to pass the test target

onurctirtir:
Maybe we can apply the suggestion from https://github.com/citusdata/test-automation/issues/242 for valgrind tests in this pr.

.circleci/config.yml
```diff
@@ -69,7 +69,7 @@ jobs:
           name: delete the given resource group
           no_output_timeout: 10m             
 
-  valgrind-test:
+  valgrind-multi-test:
```
onurctirtir, line 72: I need to drop CI jobs as we discussed with @gledis69

README.md
```diff
@@ -695,13 +701,18 @@ into the VM-s.
 On the coordinator node:
 
 ```bash
-# an example usage: Use PostgreSQL 12.1 and run valgrind test on enterprise/enterprise-master
-fab use.postgres:12.1 use.enterprise:enterprise-master run.valgrind
+# an example usage: Use PostgreSQL 14.2 and run valgrind test on enterprise/enterprise-master
+fab use.postgres:14.2 use.enterprise:enterprise-master run.valgrind
 ```
```
gledis69, line 706: Either update this with the run.valgrind:<specific_schedule> and appropriate comment
OR remove it completely, and update the README to say smth like: 
`The command can be run without tmux, but since the valgrind test run for too long we prefer to run in a detached session with tmux`.

gledis69 (approved):
LGTM

README.md
onurctirtir: > The command can be run without tmux, but since the valgrind test run for too long we prefer to run in a detached session with tmux

This is already mentioned in a few lines above:
```
However as valgrind tests take too much time to complete, we recommend you to run valgrind tests in a detached session:
```

And given that this is just an example, I usually avoid updating `12.1` but your suggestion makes sense.

PR #236, opened 2021-09-09
Increase max wal size

SaitTalhaNisanci:
With postgres 14, the `checkpoint_completion_target` is 0.9 by default and it is 0.5 with previous versions. It makes sense to use the same value for different postgres versions in our tests.

Also `max_wal_size = 1GB` is quite small for our tests and we increase it to 50GB to prevent checkpoints during tests.


onderkalaci (approved):
I think it is 50 on HSC as well, so why not :)

PR #243, opened 2022-03-17
Updating the HammerDb test automation

gledis69:
* Remove dependency from the benchmarking repo.
* Change network security policies to avoid deployment conflicts.
* HammerDb 4.4 support.
* Updated README.md
* Removed some old code.


README.md
```diff
@@ -67,6 +67,10 @@ If your subscriptions list doesn't contain `Azure SQL DB Project Orcas - CitusDa
     ssh-add
     ```
 
+5. You should setup your VPN to be able to connect to Azure VM-s if your tests are not running on CircleCI. Doing this as of lastest consists of:
```
onurctirtir, line 70: iirc circle ci does also suffer from the same problem, what should we do about that ?

Also, nit:

```suggestion
5. You should setup your VPN to be able to connect to Azure VM-s if your tests are not running on CircleCI. Doing this as of latest consists of:
```

azure/azuredeploy.json
```diff
@@ -205,7 +205,7 @@
                             "sourceAddressPrefix": "*",
                             "destinationAddressPrefix": "*",
                             "access": "Allow",
-                            "priority": 101,
+                            "priority": 102,
```
onurctirtir, line 208: Couldn't quite follow this change :) Can you note the reason behind changing the priority here and in other places ?

azure/connect.sh
```diff
@@ -21,6 +21,4 @@ ssh_port=$(az deployment group show -g ${RESOURCE_GROUP_NAME} -n azuredeploy --q
 public_ip=$(echo ${public_ip} | cut -d "\"" -f 2)
 ssh_port=$(echo ${ssh_port} | cut -d "\"" -f 2)
 
-./add-local-ip.sh
```
onurctirtir, line 24: why do we remove that ?

hammerdb/sql/tpcc-distribute-funcs.sql
```diff
@@ -1,14 +1,3 @@
-
--- THESE SHOULD BE UNCOMMENTED IF STORED PRODECURES ARE FALSE IN CONFIGS
```
onurctirtir, line 2: I guess this is because now we distribute the functions automatically, right ?

But what happens if we want to test against older versions of citus ?

azure/azuredeploy.json
gledis69: The old priorities conflict with some priorities from default network security policies that Azure deploys. The conflicts then result to failed deployments. I got these numbers by trial and error until there were no more conflicts in deployments. Depending on how the default Azure policies change these priorities might need to be changed as well in the future so I am not super happy with this hacky approach I took here.

README.md
gledis69: I am not sure why the issue happens with circle CI. The problem with Azure is due to them changing the rules of connecting to VM-s with ssh.

azure/connect.sh
gledis69: I think was some old configuration needed to allow your public ip to be connect to the VM-s. That is no longer necessary (and unreliable from my experience running the tests) since configuring and connection to the VPN should work just fine. 

This is the case for HammerDb tests. I assume this extends to the valgrind tests run on Azure VM-s as well. To make sure, it may be worth testing this.

hammerdb/sql/tpcc-distribute-funcs.sql
gledis69: I added the comment back

routes.ps1
```diff
@@ -0,0 +1,65 @@
...
+            ],
+            check=True,
+        )
+    else:
+        # TODO: do something similar for IPv6
+        ...
+"@
+echo $python_script | python
+
+Read-Host "Press Enter to continue..."
```
onurctirtir, line 65: minor: let's have a \n here :)

hammerdb/download-hammerdb.sh
```diff
@@ -0,0 +1,62 @@
...
+    echo "Expected $SHA1" 1>&2
+    exit 1
+fi
+
+if ! test -f "$OUTPUT"
+then
+    mv "$CHECK_OUTPUT" "$OUTPUT"
+fi
+
+tar -xf "$OUTPUT"
```
onurctirtir, line 62: minor: let's have a \n here :)

onurctirtir (approved):

PR #246, opened 2022-04-21
Fix "You are using an RSA key with SHA-1"

gledis69:
### Fix
The automated valgrind tests fail on checkout due to:
`ERROR: You are using an RSA key with SHA-1, which is no longer allowed.`
Azure VM-s fail with a similar error when trying to checkout the results repo.

As per the suggestion here:
https://discuss.circleci.com/t/circleci-user-key-uses-rsa-with-sha1-which-github-has-deprecated/42548

I updated the OS of:
* the valgrind tests form Ubuntu 14 to Ubuntu 20.
* the Azure VM-s from CentOS 7.2 to CentOS 7.9

There was nothing wrong with the keys themselves, it is a matter of the old libraries from the old OS-es which use SHA-1 to communicate with git.

### Note
Also changed the way ssh-keys are added since the agent used with the old method was being overwritten by CircleCI's agent. The commands being run in the CircleCI config were:

```bash
cd ./azure
./add-sshkeys
./citus-bot.sh ...
```

The add ssh key line creates an ssh-agent and picks up the key added via fingerprint. Ssh agent creation relies on variable exporting (e.g. AGENT_PORT, AGENT_ID etc.). Simply calling ./add-sshkeys does not export the variables declared inside the script unless explicitly calling `source ./add-sshkeys`. So the citus-bot script was just using an ssh-agent created by CircleCI and not the one we configure.

To fix that, I added `source ./add-sshkeys` inside citus-bot.sh, but I could also add `source` in front of the command in the yml.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,7 +14,6 @@ jobs:
       - run:
           command: |
             cd ./azure
-            ./add-sshkey.sh
             ./citus-bot.sh citusbot_pgbench_test_resource_group
           name: install dependencies and run pgbench tests
           no_output_timeout: 10m  
@@ -33,7 +32,6 @@ jobs:
       - run:
           command: |
             cd ./azure
-            ./add-sshkey.sh
             ./citus-bot.sh citusbot_scale_test_resource_group
           name: install dependencies and run scale tests
           no_output_timeout: 10m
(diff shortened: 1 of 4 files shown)


onurctirtir (approved):
let's see how it goes with the new vms

PR #249, opened 2022-05-26
Update README.md

gledis69:
The ssh-agent needs to be set before ./create-cluster.sh since the script uses the public key registered with the agent to setup the vm-s so they allow connections only from ssh connection requests which have the appropriate private key.


onurctirtir (approved):
Can you also update the non-TLDR section (if needed) :)

PR #256, opened 2022-08-25
Updates and fixes

gledis69:
Some general fixes and maintenance.
The commit nicely encapsulate the changes and describe what is being done so consider reviewing in a per commit basis.

### DIFF
--- a/azure/citus-bot.sh
+++ b/azure/citus-bot.sh
@@ -22,8 +22,6 @@ export RESOURCE_GROUP_NAME="$1"
 
 ./create-cluster.sh
 
-trap - EXIT
-
 ssh_port=$(rg_get_ssh_port ${RESOURCE_GROUP_NAME})
 public_ip=$(rg_get_public_ip ${RESOURCE_GROUP_NAME})
 
--- a/azure/citus-bot.sh
+++ b/azure/citus-bot.sh
@@ -12,30 +12,6 @@ source commons.sh
 # by CircleCI
 source ./add-sshkey.sh
 
-# ssh_execute tries to send the given command over the given connection multiple times.
-# It uses the custom ssh port 3456.
-ssh_execute() {
-   ip=$1
-   ssh_port=$2
-   shift 2;
-   command=$@
-   n=0
-   until [ $n -ge 10 ]
-   do
-      ssh -o "StrictHostKeyChecking no" -A -p "${ssh_port}" pguser@${ip} "source ~/.bash_profile;${command}" && break
-      rc=$? # get return code
-      if [ $rc -ne 255 ] ; then
-         # if the error code is not 255 we didn't get a connection error.
-         exit 1
-      fi
-      n=$[$n+1]
-   done
-
-   if [ $n == 10 ]; then
-      exit 1
-   fi
-}
-
 function cleanup {
     sh ./delete-resource-group.sh
 }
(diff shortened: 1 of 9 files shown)


fabfile/tpch_confs/tpch_default.ini
```diff
@@ -1,6 +1,6 @@
 [test]
-use_enterprise: on
-postgres_citus_versions: [('12.1', 'enterprise-master'), ('12.1', 'release-9.2')]
+use_enterprise: off
+postgres_citus_versions: [('14.3', 'release-11.0')]
```
onurctirtir, line 3: maybe we should use pg 14.5

azure/citus-bot.sh
```diff
@@ -22,8 +22,6 @@ export RESOURCE_GROUP_NAME="$1"
 
 ./create-cluster.sh
 
-trap - EXIT
-
 ssh_port=$(rg_get_ssh_port ${RESOURCE_GROUP_NAME})
```
onurctirtir, line 25: I think we shouldn't drop that since I feel like we would still want to drop the resource group if something fails around L24-L25, or shouldn't we ?

azure/create-cluster.sh
```diff
@@ -43,7 +43,7 @@ current_branch_name=$(git branch --show-current)
 export BRANCH=${CIRCLE_BRANCH:=$current_branch_name}
 
 # get local public ip 
-local_public_ip=$(curl https://ipinfo.io/ip)
+local_public_ip=$(host myip.opendns.com resolver1.opendns.com | grep "myip.opendns.com has" | awk '{print $4}')
```
onurctirtir, line 46: While the lhs one only returns the ip address, the rhs one appends an "\n" to the end, so why do we want the rhs one ?

fabfile/tpch_confs/tpch_default.ini
onurctirtir: (and two other files too)

gledis69: Also I will turn the enterprise flags off for now.
Depending on how much work might there be, I can also clean up the enterprise logic, maybe in another PR.

azure/citus-bot.sh
gledis69: The git diffs are a bit misleading for this one.
On L19 we actually trap the EXIT with the delete resource group script.
This line that I removed was overriding the previous trap with nothing, so when exiting, the resource group was not being cleaned.

azure/create-cluster.sh
gledis69: Actually, @naisila tested and this command does not work either.
MSVPN causes pinging both of these sites to timeout.
If I remember correctly, the hack to get this to work is to turn on MSVPN after running the script. Just before ssh-ing into the cluster.
At the very least the README needs to be updated, though I would prefer if we can actually fix the pinging.
Will look some more into this.

gledis69: Apparently MSVPN is not the only culprit here:
https://github.com/microsoft/WSL/issues/4517
WSL + VPN actually causes this to hang.
Networking in WSL is tricky.

azure/citus-bot.sh
onurctirtir: yeah this is true, so now I'm wondering why we were removing the trap after creating cluster, maybe it's something leftover from a debugging session :)

azure/create-cluster.sh
onurctirtir: oh really ? okay then :)

onurctirtir: I'm wondering if \n at the end could cause any issues then, but I don't think so --let's see

onurctirtir (approved):

PR #267, opened 2023-01-27
Update citus, PG and HammerDB default versions

naisila:
Drafting this since we can wait for Citus 11.2 release and replace "main" branch references with "release-11.2"


gledis69:
Yeahhhh, keeping these values constantly updated feels like a hassle. Might be worth to write a script to do a sweeping find and replace in the whole repo. bash/python should do

SaitTalhaNisanci:
Not sure if it covers what you want to do, but there is already a script updating some versions here: https://github.com/citusdata/test-automation/tree/master/release

naisila:
Yeap, I already used @SaitTalhaNisanci 's script to update most of the values, and also updated the script's description.

gledis69 (approved):

PR #270, opened 2023-03-30
Run valgrind tests on any schedule

hanefi:
Citus recently added new make targets to run valgrind tests on any regression, failure or isolation test schedule on https://github.com/citusdata/citus/pull/6667. This PR builds on top of that and adds the capability to run any schedule with valgrind enabled.

The make target name is inferred from the schedule name and it can be broken if a missing schedule name is used or the schedule name belongs to another unsupported test suite such as arbitrary config tests or PG upgrade tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ required for testing citus.
   * [Getting Started](#azure-getting-started)
     * [Setup steps for each test](#azure-setup-steps)
     * [Steps to delete a cluster](#azure-delete-cluster)
-  * [Under The Hood](#under-the-hood)  
+  * [Under The Hood](#under-the-hood)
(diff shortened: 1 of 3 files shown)


onurctirtir (approved):
Did you have a chance to run the tests at least based on i) multi-1-schedule & ii) isolation-schedule to make sure that this is working fine?

hanefi:
> Did you have a chance to run the tests at least based on i) multi-1-schedule & ii) isolation-schedule to make sure that this is working fine?

I ran base schedules for isolation failure and regression tests until completion. I started all of the schedules you mentioned but I cancelled after I saw that some of the tests have passed successfully in those schedules.

PR #278, opened 2025-01-21
Improve & simplify valgrind testing

onurctirtir:
* Reduce the dependency on the other constructs of this repo, e.g., creation & deallocation of Azure VMs via the existing tooling doesn't seem so much stable. So, introduce a simple docker image that can allow us running multiple valgrind tests in a single vm in parallel and anywhere. Also, we don't need to use `create_cluster.sh` and its friends to create an Azure VM for valgrind testing anymore because it doesn't make sense anyways. See the changes in README.md for more details.
* Get rid of valgrind code from fab infra because using this infra for valgrind testing was both unnecessary and this was heavily complicating the things. Instead, a simpler infra that we can also use in our local machines seems much more useful and easier to maintain and understand.
* Also remove fabfile/pg_report_vg_query.patch since the Postgres patch that we depend on to report the query string that causes the memory error has been merged into Postgres as of PG16.
* Automatically collect stack traces from valgrind core files that are generated in case of a process crash.

Note to reviewer - **although the PR touches 10 files, this PR mainly introduces following three files** and updates the README on the usage of valgrind testing infra. The other 6 files are only modified to remove the old infra.
* valgrind/Docker/Dockerfile
  The dockerfile to build a simple docker image that can be used to run valgrind tests.
  This has nothing but Postgres & Citus built with the flags that are needed to properly run valgrind tests and valgrind itself.
* valgrind/Docker/run_valgrind_test.sh
  The script that we run within the container to run the valgrind test for the specified test schedule.
  We could set the entrypoint of the container to this script but we keep it as a separate script for simplicity.
* run.sh
  The host side script that we use to build the container and run the valgrind tests through the container.

I strongly recommend reading [the relevant section of the README from this PR](https://github.com/citusdata/test-automation/tree/new-valgrind-runner?tab=readme-ov-file#valgrind) to better understand the changes.


valgrind/docker/Dockerfile
```diff
@@ -0,0 +1,37 @@
+FROM ubuntu:24.04
+
+RUN apt-get update && apt-get install -y git
```
thanodnl: `apt update` pulls in a lot of data that sticks around on this layer, making the image uselessly bigger. It is common to `apt clean` after you have installed something eg: https://github.com/citusdata/citus/blob/117bd1d04fe55b4001012b3923216d5df0fda10e/.devcontainer/Dockerfile#L43C5-L43C14

There are a couple of more occurances of `apt update && apt install ...` down below

valgrind/docker/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+RUN git clone https://github.com/theory/pgenv
+
+WORKDIR /pgenv
+RUN git fetch && git checkout v1.3.6b
+
+RUN apt-get update && apt-get install -y make curl
+RUN ./bin/pgenv config init
+RUN sed -i 's/^PGENV_CONFIGURE_OPTIONS=.*/PGENV_CONFIGURE_OPTIONS=([0]="--enable-debug" [1]="--enable-depend" [2]="--enable-cassert" [3]="CFLAGS=-ggdb -Og -fno-omit-frame-pointer -DUSE_VALGRIND")/' ./config/default.conf
+RUN sed -i "s/^PGENV_MAKE_OPTIONS=.*/PGENV_MAKE_OPTIONS=([0]='-sj$(nproc)')/" ./config/default.conf
```
thanodnl, line 18: It is a bit hard to read what config you are running, could you copy the config file in?
eg https://github.com/citusdata/citus/blob/117bd1d04fe55b4001012b3923216d5df0fda10e/.devcontainer/Dockerfile#L63

This adds an extra file, but makes it easier to understand what the config is, that we are using.

valgrind/docker/Dockerfile
```diff
@@ -0,0 +1,37 @@
+FROM ubuntu:24.04
+
+RUN apt-get update && apt-get install -y git
+RUN git clone https://github.com/theory/pgenv
+
+WORKDIR /pgenv
+RUN git fetch && git checkout v1.3.6b
```
thanodnl: as per https://stackoverflow.com/questions/8932389/git-shallow-clone-to-specific-tag

```suggestion
RUN git clone --branch v1.3.6b --depth 1 https://github.com/theory/pgenv
```

less code, less data, instantly in the state you want it in

valgrind/docker/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+
+RUN apt-get update && apt-get install -y make curl
+RUN ./bin/pgenv config init
+RUN sed -i 's/^PGENV_CONFIGURE_OPTIONS=.*/PGENV_CONFIGURE_OPTIONS=([0]="--enable-debug" [1]="--enable-depend" [2]="--enable-cassert" [3]="CFLAGS=-ggdb -Og -fno-omit-frame-pointer -DUSE_VALGRIND")/' ./config/default.conf
+RUN sed -i "s/^PGENV_MAKE_OPTIONS=.*/PGENV_MAKE_OPTIONS=([0]='-sj$(nproc)')/" ./config/default.conf
+
+RUN apt-get update && apt-get install -y bzip2 build-essential libicu-dev pkg-config flex bison liblz4-dev libreadline-dev libzstd-dev libxslt1-dev zlib1g-dev zlib1g-dev libxml2-utils docbook-xsl xsltproc valgrind
```
thanodnl: when installing multiple packages it is easier maintainable if you put every package on a new line and sort them. Keeping an extra ` \` at the end of the last package as well for easier diffing in the future, but that is also enforced by the apt clean you want to have in the same `RUN` as the update and install.

example https://github.com/citusdata/citus/blob/117bd1d04fe55b4001012b3923216d5df0fda10e/.devcontainer/Dockerfile#L8-L35

after each package is on a separate line use your editor to sort the packages. Future edits should remember to insert in sorted position. This will also identify double mentions of the same package, like `zlib1g-dev` in your list.

valgrind/docker/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+ARG PG_VERSION
+RUN ./bin/pgenv build $PG_VERSION && ./bin/pgenv switch $PG_VERSION
```
thanodnl, line 41: you will want to add some makeflags here to speed up the build to use all available cores.

https://github.com/citusdata/citus/blob/117bd1d04fe55b4001012b3923216d5df0fda10e/.devcontainer/Dockerfile#L71

valgrind/docker/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+WORKDIR /
+
+RUN git clone https://github.com/citusdata/citus
+WORKDIR /citus
+ARG CITUS_VERSION
+RUN git fetch && git checkout $CITUS_VERSION
```
thanodnl: You can probably use the same `--branch ... --depth 1` as above

valgrind/docker/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+RUN PATH=/pgenv/pgsql/bin/:$PATH && ./configure --without-libcurl && make -sj$(nproc) install-all
+WORKDIR /
+
+# install python3 as it's required by Citus test suite
+RUN apt-get update && apt-get install -y python3
```
thanodnl: any reason to _not_ add this to the bigger install above?

thanodnl (commented):
I left some pointers to optimize the docker image a bit. I have no real understanding of how we use valgrind, so would be good to get someone involved in day-to-day citus to checkout and try to use this change to run a valgrind test before merging - hence the lack of approval for now.

valgrind/docker/Dockerfile
thanodnl: If possible I would try to get all installs in a single run, unless there is a good reason to have them at different layers throughout the file

valgrind/run.sh
```diff
@@ -0,0 +1,63 @@
...
+docker run --name "$container_name" "$image_name" /run_valgrind_test.sh "$TEST_SCHEDULE" "$make_check_target"
+
+end_time=$(date +%s)
+
+echo "Copying artifacts from container to host"
+
+docker cp "$container_name":/citus/src/test/regress/regression.diffs "$test_artifacts_dir/"
+docker cp "$container_name":/citus/src/test/regress/regression.out "$test_artifacts_dir/"
+docker cp "$container_name":/citus/src/test/regress/citus_valgrind_test_log.txt "$test_artifacts_dir/"
+docker cp "$container_name":/citus/src/test/regress/gdb_core_backtraces "$test_artifacts_dir/"
```
thanodnl: given that we have copied out the files, would it make sense to remove the container after the run? If there is reason to keep it lets comment _why_ we are _not_ removing the container.

especially when running many schedules it pollutes the docker environment of the eng quite a bit.

onurctirtir: As readme file mentions, we keep them in case the developer wants to further investigate the coredumps.

valgrind/docker/Dockerfile
onurctirtir: Mostly keeping them separate to have a very clear list of dependencies listed for each of Postgres, Citus and the other main things.

valgrind/docker/Dockerfile
onurctirtir: https://github.com/citusdata/test-automation/pull/278#discussion_r2300302692

valgrind/docker/Dockerfile
onurctirtir: I think we're already ensuring this when writing make flags for pgenv above

valgrind/docker/Dockerfile
onurctirtir: I was mostly avoiding any extra steps to have a fast build, rather than optimizing for image size. But let me still do `apt clean` as it should not take a very long time.

naisila (approved):

